Issue #3, opened 2019-05-10
Designing the export class


tommylees112:
The requirements for the export class:
- [ ] be able to get data from multiple sources 
- [ ] be flexible (allow for downloading multiple variables without changing code) 
- [ ] base classes with behaviour specifying most user requirements

OPINIONATED on data format folder
- pipeline with full control of folder
- How to store the data?

@gabrieltseng

PR #9, opened 2019-05-10
WIP: Add CDS exporter

gabrieltseng:
Related to #3


src/exporters/__init__.py
```diff
@@ -1 +1,3 @@
 from .cds import CDSExporter
+
```
tommylees112: what do these do in the `__init__.py` files?

gabrieltseng: They make imports easier. This way, when we are running code from this library, we can do
`from src.exporters import CDSExporter` 
instead of
`from src.exporters.cds import CDSExporter`

It's especially useful when we have multiple exporters:

`from src.exporters import CDSExporter, otherExporter, ....` 
instead of
`from src.exporters.cds import CDSExporter`
`from src.exporters.other import OtherExporter`
`...`

src/exporters/cds.py
```diff
@@ -0,0 +1,66 @@
...
+                date_str = f'{years[0]}_{years[-1]}'
+            else:
+                date_str = str(years[0])
+        elif 'date' in selection_request:
+            date_str = selection_request['date'].replace('/', '_')
+
+        # force all data exports to be in netcdf format
+        # TODO: This is not possible for some exports. We should select
+        # our preferences and force those choices
+        selection_request['format'] = 'netcdf'
```
gabrieltseng: TODO: transform non-netcdf files into netcdf files as part of the export

src/exporters/cds.py
```diff
@@ -0,0 +1,66 @@
...
+        self.client = cdsapi.Client()
+
+    @staticmethod
+    def make_filename(dataset: str, selection_request: Dict) -> str:
+        """Makes the appropriate filename for a CDS export
+        """
+        date_str = ''
+        if 'year' in selection_request:
+            years = selection_request['year']
+            if len(years) > 1:
```
gabrieltseng: TODO: break up exports of more than a year and do them in chunks.

This will need to happen whether the argument is passed as `year` or `date`

src/exporters/__init__.py
gabrieltseng: We should probably think of a good way to differentiate between different exporters.
Explicitly calling the class doesn't seem like the best approach; ideally there would be a way to automatically work out which data store (and therefore which exporter) is being used

src/exporters/cds.py
tommylees112: we can use that `engine='cfgrib'` argument using `ds = xarray.open_dataset(filepath, engine='cfgrib')` as [here](https://github.com/tommylees112/esowc_notes/blob/master/notebooks/01_gt_exploration.ipynb) 

and then you can save using:
`ds.to_netcdf(output_filename)`

src/exporters/__init__.py
tommylees112: so a different exporter for what each ERA5 data-store?
- [Monthly Averaged Single Level Data](https://cds.climate.copernicus.eu/cdsapp#!/dataset/reanalysis-era5-single-levels-monthly-means?tab=overview)
- [Monthly Averaged Pressure Levels](https://cds.climate.copernicus.eu/cdsapp#!/dataset/reanalysis-era5-pressure-levels-monthly-means?tab=overview)
- [Hourly Single Levels](https://cds.climate.copernicus.eu/cdsapp#!/dataset/reanalysis-era5-single-levels)
- [Monthly Single Levels](https://cds.climate.copernicus.eu/cdsapp#!/dataset/reanalysis-era5-pressure-levels)

Or do you mean for something else?

tests/exporters/test_cds.py
```diff
@@ -0,0 +1,25 @@
+from src.exporters import CDSExporter
+
+
+class TestCDSExporter:
```
tommylees112: I'm having trouble running these tests:
```
In [2]: !pytest tests
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --flake8
  inifile: /Users/tommylees/github/ml_drought/pytest.ini
  rootdir: /Users/tommylees/github/ml_drought
```

Have you seen that before? Do i need to install flake8?

tommylees112: `conda install pytest-flake8 -c conda-forge` thanks @gabrieltseng !

src/exporters/__init__.py
gabrieltseng: We'd have one for the ERA5 data store, since it has a single API, must another one for Earth Engine (for example)

scripts/export_vhi.py
```diff
@@ -0,0 +1,33 @@
+from pathlib import Path
+
+import sys
+sys.path.append('..')
+from src.exporters import CDSExporter  # noqa
+
+
+def export_vhi_2018():
```
gabrieltseng: @tommylees112 , does this look right? I get the following:

```
2019-05-12 15:30:56,184 INFO Sending request to https://cds.climate.copernicus.eu/api/v2/resources/reanalysis-era5-single-levels
/Users/gabrieltseng/anaconda3/envs/esowc-drought/lib/python3.7/site-packages/urllib3/connectionpool.py:847: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  InsecureRequestWarning)
2019-05-12 15:30:56,591 INFO Request is queued
/Users/gabrieltseng/anaconda3/envs/esowc-drought/lib/python3.7/site-packages/urllib3/connectionpool.py:847: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  InsecureRequestWarning)
/Users/gabrieltseng/anaconda3/envs/esowc-drought/lib/python3.7/site-packages/urllib3/connectionpool.py:847: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  InsecureRequestWarning)
```
Followed by loads of insecure request warnings. I can't tell if this is expected behaviour, and if I just need to wait longer for the request to kick off.

tommylees112: I get loads of these too - there is a way to fix:

1) Add this into your imports
```
import certifi
import urllib3
http = urllib3.PoolManager(
    cert_reqs='CERT_REQUIRED',
    ca_certs=certifi.where()
)
```

2) Change the `.cdsapirc` file
- changed the verify property from 0 to 1 

Based on answers from [here](https://github.com/ecmwf/cdsapi/issues/4)

NOTE: his downloads is still not working so maybe this doesn't help...

src/exporters/__init__.py
tommylees112: Yes that makes sense. EE is great because hte data all COMES in teh same format but it does tie you to having google cloud storage OR google compute resources. 

An alternative is downloading scripts from the original data providers (e.g. NASA for VHI, MODIS etc.) but that could be even more ugly having to preprocess these separately.

@gabrieltseng  What do you think?

scripts/export_vhi.py
gabrieltseng: Adding the error for reference:

```
mars - INFO   - 20190512.194213 - Requesting any number of fields (request describes 52560)
mars - INFO   - 20190512.194213 - Calling mars on 'marscore', callback on 58915
mars - INFO   - 20190512.194213 - Callback at address 10.10.10.144, port 58915
mars - INFO   - 20190512.194215 - Mars client is on compute-0003.cdsprod.copernicus-climate.eu (10.10.10.144) 58915
mars - INFO   - 20190512.194215 - Mars server is on marscore (10.10.10.8) 36571
mars - INFO   - 20190512.194215 - Server task is 893 [mars]
mars - INFO   - 20190512.194215 - Request cost: 514,650 fields, 462.346 Gbytes online, nodes: mvr0000 mvr0001 mvr0002 mvr0003 mvr0004 mvr0005 mvr0006 mvr0007 mvr0008 [mars]
mars - ERROR  - 20190512.194243 - Mars server task finished in error
mars - ERROR  - 20190512.194243 - Assertion failed: handle_[n] != 0 in dataHandle, line 94 of /home/cds/git/mars-server/mars-server/src/dhslib/DSRetrieve.cc [mars]
mars - ERROR  - 20190512.194243 - Error code is -2
mars - ERROR  - 20190512.194243 - Request failed
mars - INFO   - 20190512.194243 - Request time:  wall: 31 sec
mars - INFO   - 20190512.194243 -   Processing in mars: wall: 30 sec
mars - INFO   - 20190512.194243 -   Visiting mars: wall: 30 sec
mars - INFO   - 20190512.194243 - Memory used: 103.04 Mbyte(s)
mars - ERROR  - 20190512.194243 - Some errors reported
```

gabrieltseng: Fixed!

scripts/export_precip.py
```diff
@@ -6,7 +6,11 @@
 
 
 def export_precip_2018():
-    data_path = Path('../data')
+    # if the working directory is alread ml_drought don't need ../data
+    if Path('.').absolute().as_posix().split('/')[-1] == 'ml_drought':
```
tommylees112: allow to run from different locations (e.g. a notebook) so you need the path to be 'data' only

Sorry i accidentally pushed the same git comnmit message

src/exporters/cds.py
tommylees112: @gabrieltseng shall we use the [`chunks` function](https://github.com/tommylees112/esowc_notes/compare/vhi_preprocess#diff-6563edbe8608e8420b07391a54e0c862R85) in a `utils.py` file somewhere to chunk up the requests and send them separately?

But we can still use the [batch_ftp_request](https://github.com/tommylees112/esowc_notes/compare/vhi_preprocess#diff-6563edbe8608e8420b07391a54e0c862R59) as a template.

Or the [`download_file_from_ftp(ftp_instance, filename, output_filename)` function](https://github.com/tommylees112/esowc_notes/compare/vhi_preprocess#diff-6563edbe8608e8420b07391a54e0c862R45)


Actually we might not need it because the max number of clients we're going to spin up is the number of years, so max `40`. 

What do you think?

tommylees112: where do we do this? This code is inside the `make_filename` function but do we not need to do this elsewhere?

tommylees112: @gabrieltseng How do you use `date`? I haven't seen it in any of the API requests?

gabrieltseng: yea, if we are going to do the `variable/year/month/file` as discussed this whole function will need to be totally rewritten.

`date_str` is used to make the filename, so that each file being downloaded is uniquely identified by its variable and date range

src/exporters/cds.py
```diff
@@ -210,13 +208,99 @@ def export(self,
...
+            path to the downloaded data
+        """
+
+        # create the default template for the selection request
+        processed_selection_request = create_selection_request(selection_request, dataset, granularity)
 
         # override with arguments passed by the user
         if selection_request is not None:
-            for key, val in selection_request.items():
-                processed_selection_request[key] = val
```
tommylees112: @gabrieltseng can we rename this `default_selection_request` and the other `user_selection_request`. I think it adds clarity

src/exporters/cds.py
```diff
@@ -215,7 +228,7 @@ def create_selection_request(self,
         # update with user arguments
         if selection_request is not None:
             for key, val in selection_request.items():
-                processed_selection_request[key] = [str(x) for x in val]
+                processed_selection_request[key] = [self._correct_input(x, key) for x in val]
```
tommylees112: Yes like this a lot great work!

src/exporters/cds.py
```diff
@@ -198,8 +198,13 @@ def _correct_input(value, key):
         return str(value)
 
     @staticmethod
-    def _check_iterable(value, key):
-        pass
+    def _check_iterable(self, value, key):
```
tommylees112: @gabrieltseng I need some help here

The alternative is:
`_check_iterable(value, key):`

Without the `self` argument in the `_check_iterable` function, testing the function [in the notebook](https://github.com/esowc/ml_drought/blob/26bf1e76ff0f3259715eadf9f42a3ae3a466d52f/notebooks/draft/01_tl_initial_test.ipynb) fails:
```
TypeError: _check_iterable() takes 2 positional arguments but 3 were given
```

So I wonder how the `@staticmethod` actually works and why pytest passes but the notebook fails?

The function is there to catch the errors that the user provides only an integer for a month for example

src/exporters/cds.py
```diff
@@ -75,38 +76,56 @@ def create_area(region: Region) -> str:
...
+    def _filename_from_selection_request(time_array: List, time: str) -> str:
+        if len(time_array) > 1:
+            if time in {'month', 'year'}:
+                warnings.warn(f'More than 1 {time} of data being exported! '
                               'Export times may be significant.')
-                years.sort()
-                date_str = f'{years[0]}_{years[-1]}'
-            else:
-                date_str = str(years[0])
-        elif 'date' in selection_request:
```
tommylees112: @gabrieltseng so is `date` the date that the user made the download? or the `dates` of the data

tests/exporters/test_cds.py
```diff
@@ -97,3 +118,35 @@ def test_user_defined_selection_requests(self, cdsapi_mock):
...
+        }
+
+        output_paths = exporter.export('precipitation',
+                                       dataset='era5',
+                                       granularity='hourly',
+                                       selection_request=user_defined_arguments,
+                                       break_up=True)
+
+        raw_folder = tmp_path / 'raw'
+        expected_paths = [
```
tommylees112: @gabrieltseng so how does this `@patch('cdsapi.Client')` `Mock` test thing actually create the .nc files? Surely these are only the result of actually calling the API?

src/exporters/cds.py
```diff
@@ -0,0 +1,309 @@
...
+        return "/".join(["{:.3f}".format(x) for x in x])
+
+    @staticmethod
+    def _filename_from_selection_request(time_array: List, time: str) -> str:
+        if len(time_array) > 1:
+            if time in {'month', 'year'}:
+                warnings.warn(f'More than 1 {time} of data being exported! '
+                              'Export times may be significant.')
+            time_array.sort()
+            time_array_str = f'{time_array[0]}_{time_array[-1]}'
```
tommylees112: @gabrieltseng another question - does it really make sense to have a separate folder for the two months?

```
.
├── README.md
└── raw
    ├── reanalysis-era5-single-levels
    │   └── precipitation
    │       ├── 1979_2019
    │       │   ├── 01_12
    │       │   └── 12
    │       ├── 1980
    │       │   ├── 09
    │       │   └── 12
    │       └── 1980_2010
    │           └── 09_12
```

How will the files be saved in folder `raw/reanalysis-era5-single-levels/precipitation/1980_2010/09_12`? 

Because currently they are written as: `f'{days}.nc'` but isn't that going to lead to a whole load of files overwriting each other because you've got every file from `1980 - 2010` and every month from `September - December` all saved by their `day` names ?

tests/exporters/test_cds.py
drop-gabriel: No `.nc` files are actually created. The test only checks that the correct filenames were created (broken up by month). If those filenames are made, the correct API requests should have been made (also broken up by months), since, the selection request is used to make the filename.

gabrieltseng: No .nc files are actually created. The test only checks that the correct filenames were created (broken up by month). If those filenames are made, the correct API requests should have been made (also broken up by months), since, the selection request is used to make the filename.

tommylees112:
First big merge! Woohooo

PR #12, opened 2019-05-15
WIP: VHI export

gabrieltseng:
Related to #3 . An export class to download the vegetation health index.


src/exporters/vhi.py
```diff
@@ -0,0 +1,172 @@
...
+            ftp.retrlines("LIST", listing.append)
+            # extract the filename
+            filepaths = [f.split(' ')[-1] for f in listing]
+            # extract only the filenames of interest
+            vhi_files = [f for f in filepaths if ".VH.nc" in f]
+
+        return vhi_files
+
+    @staticmethod
+    def chunks(l, n):
```
tommylees112: @gabrieltseng this function appears to work (no problems my side with the code). I am not *certain* that it's parallelized but i'm 90% sure because the order that the files return is not sequential and the list is sequential when you call the FTP server. So I'm pretty sure it's working as expected.

BUT it's failing the mypy checks:
```
src/exporters/vhi.py:152: error: Argument 1 to "len" has incompatible type "int"; expected "Sized"
src/exporters/vhi.py:154: error: Value of type "int" is not indexable
```

src/exporters/vhi.py
```diff
@@ -0,0 +1,172 @@
...
+        with open(self.raw_folder / 'vhi_export_errors.pkl','wb') as f:
+            pickle.dump([error[-1] for error in outputs if error != None], f)
+
+        return
+
+    def run_parallel(vhi_files: List) -> List:
+        pool = Pool(processes=100)
+
+        # split the filenames into batches of 100
+        batches = [batch for batch in chunks(vhi_files,100)]
```
tommylees112: this is also working except for a type incompatbility.

```
src/exporters/vhi.py:172: error: Argument 1 to "chunks" of "VHIExporter" has incompatible type "List[Any]"; expected "int"
```

tommylees112: Oh it's all fixed by changing to : 
`def chunks(self, l: List, n: int) -> Generator:`

src/exporters/vhi.py
tommylees112: Oh it's all fixed by changing to : 
`def chunks(self, l: List, n: int) -> Generator:`

tommylees112:
Todo:
- [ ] check that the parallel code is actually running 
- [ ] check that reading the pickled error messages actually opens 
- [ ] check that the pickled errors are written

tommylees112:
@gabrieltseng 
FOR TODO 1:
```
- [ ] check that the parallel code is actually running
```
could we download a few files from different years just to test that the FTP request has been successful? Or does this introduce too much extra data space/requirement for travis etc.

src/exporters/vhi.py
```diff
@@ -0,0 +1,220 @@
...
+
+        # save the filenames that failed to a pickle object
+        with open(self.raw_folder / 'vhi_export_errors.pkl', 'wb') as f:
+            pickle.dump([error[-1] for error in outputs if error is not None], f)
+
+        return
+
+    def run_parallel(self,
+                     vhi_files: List,
+                     years: List[int] = [yr for yr in np.arange(1981, 2020)]
```
gabrieltseng: We shouldn't have mutable values as default arguments. One option is to have a function which creates default arguments, like the ERA5 exporter's [`get_era5_times`](https://github.com/esowc/ml_drought/blob/master/src/exporters/cds.py#L129)

src/exporters/vhi.py
```diff
@@ -0,0 +1,220 @@
...
+        ftp.cwd('/pub/corp/scsb/wguo/data/Blended_VH_4km/VH/')
+
+        # download each filename using this FTP object
+        for raw_filename in legit_filenames:
+            output_filename = (
+                make_filename(raw_folder, raw_filename, dataset='vhi')
+            )
+            download_file_from_ftp(ftp, raw_filename, output_filename)
+
+    return
```
gabrieltseng: These blank return statements aren't necessary

src/exporters/vhi.py
```diff
@@ -0,0 +1,220 @@
...
+            listing: List = []
+            ftp.retrlines("LIST", listing.append)
+            # extract the filename
+            filepaths = [f.split(' ')[-1] for f in listing]
+            # extract only the filenames of interest
+            vhi_files = [f for f in filepaths if ".VH.nc" in f]
+
+        return vhi_files
+
+    # @staticmethod
```
gabrieltseng: This can be a static method

src/exporters/vhi.py
```diff
@@ -0,0 +1,220 @@
...
+        # passing mutliple args using partial https://stackoverflow.com/a/5442981/9940782
+        args = dict(raw_folder=self.raw_folder, years=years)
+        outputs = pool.map(partial(batch_ftp_request, args), batches)
+
+        # write the output (TODO: turn into logging behaviour)
+        print("\n\n*************************")
+        print("VHI Data Downloaded")
+        print("*************************")
+        print("Errors:")
+        pprint([error for error in outputs if error is not None])
```
gabrieltseng: Since the errors are also printed in `save_errors`, they would get printed twice. Can we eliminate one of the print statements?

src/exporters/vhi.py
```diff
@@ -0,0 +1,220 @@
...
+        # save errors
+        self.save_errors(outputs)
+
+        return batches
+
+    def check_failures(self) -> List:
```
gabrieltseng: This function is never called. Can we get rid of it?

src/exporters/vhi.py
```diff
@@ -0,0 +1,220 @@
+"""
+functionality:
+1) download VHI data in parallel
+2)
+
+CDS functions:
```
gabrieltseng: Nit: there are lots of comments in this file which could be removed.

src/exporters/vhi.py
gabrieltseng: They appear in some other functions too

src/exporters/vhi.py
```diff
@@ -0,0 +1,220 @@
...
+
+        with open(pickled_error_fname, 'rb') as f:
+            errors = pickle.load(f)
+        print("*************************")
+        print("Errors:")
+        pprint(errors)
+
+        return errors
+
+    def export(self) -> List:
```
gabrieltseng: Could we add the following to this function:

- The option to change the arguments of the download (right now, I can't limit the number of years I download by calling `export`)
- Add a docstring, since this is the function users will use to interact with the exporter

tests/exporters/test_vhi.py
```diff
@@ -0,0 +1,68 @@
+from pathlib import Path
+# from unittest.mock import patch, Mock
```
gabrieltseng: Nit: comment out code to be removed

tests/exporters/test_vhi.py
```diff
@@ -0,0 +1,68 @@
...
+import pytest
+
+from src.exporters.vhi import (
+    VHIExporter,
+    _parse_time_from_filename,
+    make_filename,
+    # download_file_from_ftp
```
gabrieltseng: Same; commented out code to be removed

tests/exporters/test_vhi.py
```diff
@@ -0,0 +1,68 @@
...
+# get the root project directory
+project_dir = Path(__file__).resolve().parents[2]
+# append to sys
+sys.path.append(project_dir.as_posix())
+
+
+class TestVHIExporter:
+
+    def test_initialisation(self):
+        VHIExporter()
```
gabrieltseng: What does this test?

Since the VHI initialization just does what the base exporter does, and that is tested in `test_base.py`, this test isn't necessary

src/exporters/vhi.py
```diff
@@ -0,0 +1,220 @@
...
+    return
+
+
+class VHIExporter(BaseExporter):
+    """Exports Vegetation Health Index from NASA site
+
+    ftp.star.nesdis.noaa.gov
+    """
+
+    def __init__(self, data_folder: Path = Path('data')) -> None:
```
gabrieltseng: Since this init function only calls the BaseExporter's init function, its not necessary

tests/exporters/test_vhi.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def test_filename_illegitimate(self):
+        with pytest.raises(Exception):
+            make_filename(
+                raw_folder=project_dir / "data",
+                raw_filename='this/should/fail.nc',
+                dataset='vhi',
+            )
+
+        return
```
gabrieltseng: Same as above; empty returns are unnecessary

tests/exporters/test_vhi.py
```diff
@@ -0,0 +1,68 @@
...
+
+class TestVHIExporter:
+
+    def test_initialisation(self):
+        VHIExporter()
+        return
+
+    def test_filename_illegitimate(self):
+        with pytest.raises(Exception):
+            make_filename(
```
gabrieltseng: Nice! It would be a good idea to check too that the error message is what you expect, to make sure you aren't catching some other unexpected error

tests/exporters/test_vhi.py
```diff
@@ -0,0 +1,68 @@
...
+                raw_folder=project_dir / "data",
+                raw_filename='this/should/fail.nc',
+                dataset='vhi',
+            )
+
+        return
+
+    def test_filenames_found(self):
```
gabrieltseng: As we discussed, I would prefer if we didn't actually call the NOAA website to check this code works.

One way to avoid this is by patching ftplib: https://stackoverflow.com/questions/29817963/mocking-ftplib-ftp-for-unit-testing-python-code

.gitignore
```diff
@@ -21,3 +21,6 @@ data/*.zip*
 *.idx
 
 .ipynb_checkpoints
+
+*.out
+data/raw/*
```
gabrieltseng: Thanks :) well caught

src/exporters/vhi.py
gabrieltseng: Also, in your notebook you have a nice assert statement which makes sure the years passed are within the accepted range. It would make sense for that to be within this export function too

tests/exporters/test_vhi.py
```diff
@@ -0,0 +1,68 @@
...
+
+from src.exporters.vhi import (
+    VHIExporter,
+    _parse_time_from_filename,
+    make_filename,
+    # download_file_from_ftp
+)
+
+import sys
+# get the root project directory
```
gabrieltseng: What is this for?

src/exporters/vhi.py
tommylees112: I haven't found out where to call it yet - I think that it's still useful because if we want to 'retry' running for the files that have been missed. For example if you run out of space, it will catch the files htat haven't been downloaded and then you can clear space and continue the download?

What do you think?

gabrieltseng: Hmm, that is fair, but the errors will be printed before they are saved. I guess my concern is that having it in this class makes it unlikely users will use it.

Perhaps an alternative is to save it as a text file (so its easier to access), and then (if errors are saved) print at the end of the export `"errors saved in errors.txt"`.

This does highlight the need for logging infrastructure, which can be one of the next things we work on.

tests/exporters/test_vhi.py
tommylees112: you're right it's gone!

tests/exporters/test_vhi.py
tommylees112: My test function

```
    @patch('ftplib.FTP', autospec=True)
    def test_filenames_found(self, mock_ftp_constructor):
        mock_ftp = mock_ftp_constructor.return_value
        vhi = VHIExporter()
        fnames = vhi.get_ftp_filenames()

        mock_ftp_constructor.assert_called_with('ftp.star.nesdis.noaa.gov')
        self.assertTrue(mock_ftp.login.called)

```

My real function:

```
    @staticmethod
    def get_ftp_filenames() -> List:
        """  get the filenames containing VHI """
        with FTP('ftp.star.nesdis.noaa.gov') as ftp:
            ftp.login()
            ftp.cwd('/pub/corp/scsb/wguo/data/Blended_VH_4km/VH/')

            # append the filenames to a list
            listing: List = []
            ftp.retrlines("LIST", listing.append)
            # extract the filename
            filepaths = [f.split(' ')[-1] for f in listing]
            # extract only the filenames of interest
            vhi_files = [f for f in filepaths if ".VH.nc" in f]

        return vhi_files

```

Do I need to change the function to separate teh

tommylees112: Error message

```
tests/exporters/test_vhi.py:38:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

_mock_self = <MagicMock name='FTP' spec='FTP' id='4626376352'>, args = ('ftp.star.nesdis.noaa.gov',)
kwargs = {}, expected = "FTP('ftp.star.nesdis.noaa.gov')"

    def assert_called_with(_mock_self, *args, **kwargs):
        """assert that the mock was called with the specified arguments.

        Raises an AssertionError if the args and keyword args passed in are
        different to the last call to the mock."""
        self = _mock_self
        if self.call_args is None:
            expected = self._format_mock_call_signature(args, kwargs)
>           raise AssertionError('Expected call: %s\nNot called' % (expected,))
E           AssertionError: Expected call: FTP('ftp.star.nesdis.noaa.gov')
E           Not called

../../miniconda3/envs/crop/lib/python3.7/unittest/mock.py:811: AssertionError
```

drop-gabriel: I don't think it needs to be split into two functions. So one thing you could do is mock the `ftp.retrlines` function so that it returns a bunch of filenames (since what we want to test is that we only retrieve the `.VH.nc` filenames).

Mock has a `side_effect` feature (see this [blog post](http://fgimian.github.io/blog/2014/04/10/using-the-python-mock-library-to-fake-regular-functions-during-tests/)) which allows you to replace a function with another, so I would do that for `ftp.retrlines`, something like this:

```python
def fake_retrlines(some_string, callback):
    # I don't know if the string argument is meaningful; if it is it can be considered
    # here too
    fake_data = ['a', 'b', 'c.VI.nc', 'd', 'e.VI.nc']
    for filename in fake_data:
        callback(filename)
```

And then, within the function, add:
`@mock.patch('ftp.retrlines', side_effect=fake_retrlines)`

gabrieltseng: I don't think it needs to be split into two functions. So one thing you could do is mock the ftp.retrlines function so that it returns a bunch of filenames (since what we want to test is that we only retrieve the .VH.nc filenames).

Mock has a side_effect feature (see this blog post) which allows you to replace a function with another, so I would do that for ftp.retrlines, something like this:

```python
def fake_retrlines(some_string, callback):
    # I don't know if the string argument is meaningful; if it is it can be considered
    # here too
    fake_data = ['a', 'b', 'c.VI.nc', 'd', 'e.VI.nc']
    for filename in fake_data:
        callback(filename)
```
And then, within the function, add: `@mock.patch('ftp.retrlines', side_effect=fake_retrlines)`

tommylees112: https://stackoverflow.com/questions/28850070/python-mocking-a-context-manager?noredirect=1&lq=1

tommylees112: https://stackoverflow.com/questions/29817963/mocking-ftplib-ftp-for-unit-testing-python-code

tommylees112: Attempt 2:
```
    ```@patch('ftplib.FTP', autospec=True)
    def test_filenames_found(self, mock_ftp):
        mock_ftp.return_value = Mock()
        mock_ftp_obj = mock_ftp()

        fnames = VHIExporter.get_ftp_filenames()
        mock_ftp.assert_called_with('ftp.star.nesdis.noaa.gov')```
```

tommylees112: attempt 3:
```
    ```def test_filenames_found(self, mock_ftp):
        mock_ftp.return_value.__enter__.return_value.name = Mock()
        mock_ftp_constructor = Mock()
        vhi = VHIExporter()
        fnames = vhi.get_ftp_filenames()

        mock_ftp.assert_called_once_with('ftp.star.nesdis.noaa.gov')```
```

tests/exporters/test_vhi.py
tommylees112:
```
with pytest.raises(Exception) as e:
    make_filename(
        raw_folder=project_dir / "data",
        raw_filename='this/should/fail.nc',
        dataset='vhi',
    )
    e.match(r'filename cannot have subdirectories*')
```

tests/exporters/test_vhi.py
tommylees112: think i've fixed all of these

src/exporters/vhi.py
tommylees112: can we keep it for future requirements (extending it to run the preprocessing by default to reduce file size for example)

or just revive it when required

tests/exporters/test_vhi.py
tommylees112: good point well made

tests/exporters/test_vhi.py
tommylees112: gone!

tests/exporters/test_vhi.py
tommylees112: gone!

src/exporters/vhi.py
tommylees112: which notebook?

src/exporters/vhi.py
tommylees112: done thanks gabi!

src/exporters/vhi.py
tommylees112: One of these is returning the errors from the pickled file to the user? maybe i can remove the print statement in that function

tommylees112: done

src/exporters/vhi.py
tommylees112: changed the print statement slightly can you check in the most up to date `vhi.py` argument?

tommylees112: Will remove if you think is unnecessary

src/exporters/vhi.py
tommylees112: think i've removed alot of them

tests/exporters/test_vhi.py
tommylees112: @gabrieltseng's answer:
```
    @patch('ftplib.FTP', autospec=True)
    def test_ftp_called(self, mock_ftp):
        _ = VHIExporter.get_ftp_filenames(years=list(range(1981, 2020)))
        mock_ftp.assert_called_once_with('ftp.star.nesdis.noaa.gov')

```

src/exporters/vhi.py
tommylees112: hmm not sure i've understood this - can you check if this fixed in the new push?

src/exporters/vhi.py
tommylees112: Done!

src/exporters/vhi.py
tommylees112: okay fixed

src/exporters/vhi.py
tommylees112: nah unnecessary code good call

src/exporters/vhi.py
```diff
@@ -196,25 +167,38 @@ def check_failures(self) -> List:
...
-    def export(self) -> List:
-        """Export VHI data from the ftp server
+    @staticmethod
+    def get_default_years():
+        """ returns the default arguments for no. years """
+        years = [yr for yr in range(1981, 2020)]
+
+        return years
+
+    def export(self, years: List = [yr for yr in range(1981, 2020)]) -> List:
```
tommylees112: @gabrieltseng  does this still violate the problems here: https://github.com/esowc/ml_drought/pull/12#discussion_r285190313

If so does hte funciton above fix that?

Do we need to have 
`min_year` and `max_year` as arguments instead?

src/exporters/vhi.py
```diff
@@ -196,25 +167,38 @@ def check_failures(self) -> List:
...
+        By default write output to raw/vhi/{YEAR}/{filename}
 
-        by default write output to raw/vhi/{YEAR}/{filename}
-        self.data_folder
-        self.raw_folder
+        Arguments:
+        ---------
+        years : List
+            list of years that you want to download. Default `range(1981,2020)`
+
```
tommylees112: @gabrieltseng is the docstring okay?

https://github.com/esowc/ml_drought/pull/12#discussion_r285193949

gabrieltseng: Yep! This is great, thanks :)

pipeline_config/default.json
```diff
@@ -1,14 +1,24 @@
...
-      ],"vhi":[
-          {
-              "years":["1981"]
-          }
-      ]
-   }
+    "data": "data",
+    "export": {
+        "vhi": [
+            {"years": [2010,
```
tommylees112: @gabrieltseng Just a subset of the years for the default?

src/exporters/vhi.py
gabrieltseng: yes thank you!

pipeline_config/default.json
gabrieltseng: (To be revisited in another PR)

PR #13, opened 2019-05-15
Add entrypoint to the pipeline, and default arguments

gabrieltseng:
- [x] Add a `pipeline_config` folder where configurations from different pipelines can be run
- [x] Add a `default.json` file of default configurations, which are overriden by user-provided configurations
- [x] Add an entry point to the pipeline
- [x] Add tests (some added)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,3 +15,4 @@ RUN conda env create -f ./environment.ubuntu.cpu.yml
 COPY src ./src
 COPY tests ./tests
 COPY *.ini ./
+COPY pipeline_config ./pipeline_config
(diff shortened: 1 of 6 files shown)


tommylees112:
Great work Gabi!!

gabrieltseng:
Potential to dos, in a future PR:

- [ ] Replace arparse with python fire, when more arguments are added
- [ ] In addition to the default configuration, add a 'minimal' configuration for running simple models without downloading lots of data
- [ ] (Maybe) a more thorough check the input is as we expect it, perhaps by comparing to the defaults

PR #14, opened 2019-05-19
Vhi preprocess #preprocess

tommylees112:
TODO:
- [x] Implement a VHI cleaner to turn the raw `.nc` files into one `.nc` file subset for kenya across multiple timesteps
- [ ] Make the functions as general as possible
- [ ] WRITE TESTS

What we want:
- a preprocessing script that turns the raw VHI files into files with `latitude` `longitude` `time` dimensions
- Subset Kenya (same bounding box as the ERA5 data)


tommylees112:
Recent additions have achieved a number of things (sorry for not splitting up properly)

1) Added the datasets wanted to a [notebook](https://github.com/esowc/ml_drought/blob/vhi_preprocess/notebooks/draft/04_tl_datasets_wanted.ipynb) 
2) Added the functionality to [rerun the missing files](https://github.com/esowc/ml_drought/blob/f685433d637e172adc6d4b933678f363a0d7dae5/src/exporters/vhi.py#L211)
3) Implemented [VHI preprocessing in parallel](https://github.com/esowc/ml_drought/blob/vhi_preprocess/src/preprocess/vhi.py) which currently takes the data that is already downloaded (from the exporter) and then turns it into a subset of the data for `kenya` with `latitude` `longitude` and `time` dimensions added to the `.nc` file

src/preprocess/vhi.py
```diff
@@ -0,0 +1,173 @@
...
+import xarray as xr
+import multiprocessing
+from typing import List, Optional, Tuple, Union
+import pickle
+from functools import partial
+
+from xarray import Dataset
+
+from .base import (BasePreProcessor, get_kenya)
+from .preprocess_vhi import (
```
gabrieltseng: Since we have two files dedicated to `vhi` preprocessing, can we move them into their own folder?

gabrieltseng: (Or just make them part of the preprocessing class)

src/preprocess/preprocess_vhi.py
```diff
@@ -0,0 +1,192 @@
...
+
+from xarray import Dataset, DataArray
+from pandas._libs.tslibs.timestamps import Timestamp
+
+# ------------------------------------------------------------------------------
+# 1. assign time stamp
+# ------------------------------------------------------------------------------
+
+
+def extract_timestamp(ds: Dataset,
```
gabrieltseng: What is the motivation for keeping these functions in a seperate file? Can they be part of the `VHIPreprocessor` class?

src/preprocess/vhi.py
```diff
@@ -0,0 +1,173 @@
...
+        if not self.vhi_interim.exists():
+            self.vhi_interim.mkdir()
+
+    def get_vhi_filepaths(self) -> List[Path]:
+        return [f for f in (self.raw_folder / "vhi").glob('*/*.nc')]
+
+    @staticmethod
+    def preprocess_VHI_data(netcdf_filepath: str,
+                            output_dir: str,
+                            subset: str = 'kenya',) -> Path:
```
gabrieltseng: Since right now, `subset` has to be 'kenya' to do anything, can we take it out of the function arguments? Or make it a boolean argument, e.g.
`subset_kenya: bool = True`

src/run.py
```diff
@@ -73,7 +74,7 @@ def export(self, export_args: Dict) -> None:
             exporter = dataset2exporter[dataset](self.data)
 
             for variable in variables:
-                _ = exporter.export(**variable)
+                _ = exporter.export(**variable)  # type: ignore
 
     def run(self, config: DictWithDefaults) -> None:
```
gabrieltseng: We should add a processing function here so that it can be run as part of the pipeline.

It should also be added to the configuration files

tests/preprocess/test_vhi.py
```diff
@@ -0,0 +1,151 @@
...
+    )
+    raw_ds.to_netcdf(netcdf_filepath)
+
+    # run the preprocessing steps
+    out = v.preprocess_VHI_data(
+        netcdf_filepath.as_posix(), v.vhi_interim.as_posix(),
+    )
+    expected = "STAR_VHP.G04.C07.NC_1981_8_31_kenya_VH.nc"
+
+    assert "kenya_VH.nc" in out.name, f"Expected to find kenya_VH.nc \
```
gabrieltseng: Is this assert statement redundant given the one below?

src/preprocess/base.py
```diff
@@ -0,0 +1,23 @@
+from pathlib import Path
+
+from ..utils import Region, get_kenya
+
+
+__all__ = ['BasePreProcessor', 'Region', 'get_kenya']
+
+
+class BasePreProcessor:
+    """Base for all pre-processor classes
```
gabrieltseng: Can we better document what the processing classes are supposed to be doing here?

e.g `Processing classes are responsible for normalizing dimensions, optionally subsetting Kenya, etc..`

src/preprocess/preprocess_utils.py
```diff
@@ -0,0 +1,48 @@
+import xarray as xr
```
gabrieltseng: This file can be renamed to `utils.py`, since its already in the preprocessing folder

src/preprocess/vhi.py
```diff
@@ -0,0 +1,173 @@
...
+
+        self.vhi_interim = self.interim_folder / "vhi"
+        if not self.vhi_interim.exists():
+            self.vhi_interim.mkdir()
+
+    def get_vhi_filepaths(self) -> List[Path]:
+        return [f for f in (self.raw_folder / "vhi").glob('*/*.nc')]
+
+    @staticmethod
+    def preprocess_VHI_data(netcdf_filepath: str,
```
gabrieltseng: nit: function names should be lowercase (this would also be more consistent with the function above)

src/preprocess/vhi.py
```diff
@@ -0,0 +1,173 @@
...
+
+        # save the merged filepath
+        ds.to_netcdf(outpath)
+        print(f"Timesteps merged and saved: {outpath}")
+
+        # turn from a dask mfDataset to a Dataset (how is this done?)
+
+        return ds
+
+    def preprocess(self, subset: Optional[str] = 'kenya') -> None:
```
gabrieltseng: Same comment as above re: making the subset boolean.

src/preprocess/vhi.py
```diff
@@ -0,0 +1,173 @@
...
+        # 6. create the filepath and save to that location
+        filename = create_filename(
+            timestamp,
+            netcdf_filepath,
+            subset=True,
+            subset_name=subset
+        )
+        print(f"Saving to {output_dir}/{filename}")
+        # TODO: change to pathlib.Path objects
+        kenya_ds.to_netcdf(f"{output_dir}/{filename}")
```
gabrieltseng: Since `kenya_ds` is only defined if a subset is defined, this function will fail if a subset isn't defined

src/preprocess/vhi.py
```diff
@@ -0,0 +1,173 @@
...
+from .preprocess_vhi import (
+    extract_timestamp,
+    create_lat_lon_vectors,
+    create_new_dataset,
+    create_filename,
+)
+from .preprocess_utils import select_bounding_box_xarray
+
+
+class VHIPreprocessor(BasePreProcessor):
```
gabrieltseng: Can you add this to the `__init__` file too, so that it can be imported as

`from src.preprocess import VHIPreprocessor`

tests/preprocess/test_vhi.py
tommylees112: yes sorry had written both separately and forgot to remove

src/run.py
tommylees112: Done!

src/preprocess/base.py
tommylees112: done!! thanks mate :)

PR #16, opened 2019-05-20
Chirps Export Functionality

tommylees112:
add chirps exporting functionality, parallelised (but untested) because I had to run to the gym

TODO:
- [x] Add tests
- [ ] Test the parallel functionality works INSIDE the class (have a sneaky suspicion we might have difficulties)

This will allow us to have a precip dataset for the whole time series of interest to move onto the next stage of the processing


tommylees112:
I should add it works from the tests i ran on the [esowc_notes branch](https://github.com/tommylees112/esowc_notes/blob/vhi_preprocess/example_scripts/download_chirps.py)

gabrieltseng:
Is it necessary to parallelize this? There are only ~20 files to download

gabrieltseng:
@tommylees112 , assuming the multiprocessing works (adding a test for this really clogged things up, so I left it out), this should be good to go if it looks good to you.

Only one caveat: the environment files need to be updated for the tests to pass, but I don't want to do that until the Planet OS stuff is merged, because that also updated the environment files.

tests/exporters/test_chirps.py
```diff
@@ -1,8 +1,36 @@
...
+
+    def test_default_years(self):
+
+        expected_default_years = list(range(1981, 2020))
+
+        assert CHIRPSExporter.get_default_years() == expected_default_years, \
+            f'Expected default years to be {expected_default_years}, ' \
+            f'got {CHIRPSExporter.get_default_years()}'
+
+    @patch('os.system', autospec=True)
```
tommylees112: @gabrieltseng this is class thankyou!!

tests/exporters/test_chirps.py
```diff
@@ -30,11 +34,56 @@ def test_checkpointing(self, mock_system, tmp_path, capsys):
...
+                            '2.0.1981.02.5.tif.gz\r\n'
+
+        expected_filenames = ['chirps-v2.0.1981.01.1.tif.gz', 'chirps-v2.0.1981.01.2.tif.gz',
+                              'chirps-v2.0.1981.01.3.tif.gz', 'chirps-v2.0.1981.01.4.tif.gz',
+                              'chirps-v2.0.1981.01.5.tif.gz', 'chirps-v2.0.1981.01.6.tif.gz',
+                              'chirps-v2.0.1981.02.1.tif.gz', 'chirps-v2.0.1981.02.2.tif.gz',
+                              'chirps-v2.0.1981.02.3.tif.gz', 'chirps-v2.0.1981.02.4.tif.gz',
+                              'chirps-v2.0.1981.02.5.tif.gz']
+
+        def mockreturn(request):
```
tommylees112: @gabrieltseng this is far out mate! Can you explain to me how it works?

gabrieltseng: The functionality I was testing for was that given a string (the response body of the africa url), the code would correctly find the correct filenames.

To do this, I used [`monkeypatch`](https://stackoverflow.com/questions/5626193/what-is-monkey-patching), which replaces what a function normally does with whatever you want it to do.

I wanted this line: `response = urllib.request.urlopen(req)` to return a class which could be `.read()` to get the first ~1000 characters of the string, instead of opening a connection with the url, so I replaced its original functionality with `mockreturn`

tommylees112: niiiice! does it matter what you call the class:
```
def mockreturn(request):
    class OpenURL:
        def read(self):
            return expected_response
    open_url = OpenURL()
    return open_url
```

Also does the following therefore mean that any calls to `urllib.request.urlopen` are replaced with mockreturn?

```
monkeypatch.setattr(urllib.request, 'urlopen', mockreturn)
exporter = CHIRPSExporter(tmp_path)
filenames = exporter.get_chirps_filenames(region='africa')
```


```

gabrieltseng: The class can be called anything, and yes any calls to `urllib.request.urlopen` are replaced with mockreturn.

PR #17, opened 2019-05-21
Improve documentation

gabrieltseng:
- Update pipeline configuration files to have a `full`, which downloads all the data, and a default `minimal` which only downloads the last two full years of data (2017, 2018)
- Update main readme to include information about the pipeline entry point
- Add a readme in the pipeline configuration folder which goes into more detail about the pipeline

- Remove the todo section from the main readme


tommylees112:
Thanks @gabrieltseng this is super helpful!!

PR #22, opened 2019-05-27
global default & unique filepath

tommylees112:
update the CHIRPS downloader to have GLOBAL by default (netcdf files) and to create a unique path for these two options


gabrieltseng:
@tommylees112 , I changed it so that the region and chirps folder are separately defined. Otherwise, if export were to be called twice on an already initialized exporter (e.g. once to pull `africa` data, once to pull `global` data), then the first export would save to `chirps_folder / africa`, but the second would save to `chirps_folder / africa / global`.

If it looks good to you, it can be merged.

PR #21, opened 2019-05-27
WIP: Preprocessing

gabrieltseng:
Preprocessing the CHIRPS dataset (and other datasets so they play well with CHIRPS)

- [x] General spatial regridding functionality
- [x] General temporal regridding functionality
- [x] Tests for everything

PR #25, opened 2019-05-30
INVERT the HEIGHT because downloaded upside down

tommylees112:
See this fix [here](https://github.com/esowc/ml_drought/pull/21#issuecomment-497149534)


tommylees112:
This is not a fix :(

PR #27, opened 2019-05-31
Exporters/seas5

tommylees112:
Setting up the seasonal exporter data


src/exporters/cds.py
```diff
@@ -280,23 +286,31 @@ def export(self,
...
 
-                # multiprocessing of the paths
                 output_paths.append(
-                    p.apply_async(
-                        self._export,
-                        args=(dataset, updated_request, show_api_request))
+                    self._export(dataset, updated_request, show_api_request)
                 )
+
+                if parallel:  # Run in parallel
```
tommylees112: there are so many of these `if parallel` statements - might be better to have a separate function for calling it in parallel (just for ease of reading the code). What do you think?

gabrieltseng: Hmm given that the functions would do the exact same thing (and require the same input arguments), just one in parallel and one not, I would prefer to keep it all together, rather than expect the user to remember two different function names.

src/exporters/cds.py
```diff
@@ -241,7 +242,9 @@ def export(self,
                granularity: str = 'hourly',
                show_api_request: bool = True,
                selection_request: Optional[Dict] = None,
-               break_up: bool = True) -> List[Path]:
+               break_up: bool = True,
+               parallel: bool = True,
+               N_parallel_requests: int = 3) -> List[Path]:
```
gabrieltseng: @tommylees112 , an option here is to not have the boolean parallel field, and only use the n_parallel_requests.
If n_parallel_requests = 1, then it wouldn't run in parallel

src/exporters/s5.py
```diff
@@ -0,0 +1,484 @@
...
+        ):
+            updated_request = processed_selection_request.copy()
+            updated_request["year"] = [year]
+            updated_request["month"] = [month]
+
+            if N_parallel_requests > 1:  # Run in parallel
+                # multiprocessing of the paths
+                in_parallel = True
+                output_paths.append(
+                    p.apply_async(
```
tommylees112: @gabrieltseng i need some help with this. I tried to use the `from pathos.pools import _ThreadPool as pool` object because it was overcoming the `pickling` error which i was having with the `multiprocessing.pool`.

The issue is that I want to just spin up a load of these in parallel (which works with the bash/sed approach). I wonder if it is something to do with trying to do parallel tasks inside a . class?

Anyway any ideas would be super grateful for!

tests/exporters/test_s5.py
```diff
@@ -0,0 +1,299 @@
...
+                granularity=granularities[ix],
+                pressure_level=pressure_levels[ix],
+            )
+
+            assert (
+                s5.dataset == expected
+            ), f"\
+            Expected: {expected}. Got: {s5.dataset}"
+
+    @patch("cdsapi.Client")
```
tommylees112: @gabrieltseng could you possibly help with this pytest patching malarkey? unfortunatley it's super tricky because my pytest is broken - but i ran these tests as simple assert statements in my ipython console. i really need to get stuck into an IDE

src/exporters/s5.py
```diff
@@ -0,0 +1,488 @@
...
+        self.pressure_level = pressure_level
+        self.granularity = granularity
+
+        if dataset is None:
+            self.dataset = self.get_dataset(self.granularity, self.pressure_level)
+        else:
+            self.dataset = dataset
+
+        # get the reference dictionary that corresponds to that dataset
+        self.dataset_reference = dataset_reference[self.dataset]
```
tommylees112: @gabrieltseng why does this assume that the type is `Optional[Sequence[str]]` when 1) it's not optional 2) it's a dictionary 3) the values for the dictionary vary between `List`, `str`, `None`

```
src/exporters/s5.py:359: error: Incompatible return value type (got "Optional[Sequence[str]]", expected "List[Any]")
src/exporters/s5.py:388: error: Unsupported right operand type for in ("Optional[Sequence[str]]")
```

src/exporters/s5.py
```diff
@@ -0,0 +1,488 @@
...
+            "system": "5",
+            "variable": [variable],
+        }
+
+        # add product_type if required
+        if self.product_type is not None:
+            processed_selection_request.update({"product_type": [self.product_type]})
+
+        # get the initialisation time information
+        init_times_dict = self.get_s5_initialisation_times(
```
tommylees112: ???????? @gabrieltseng 

How do I specify that these are `Optional` without mypy then complaining that I am comparing `None` type to 
```
src/exporters/s5.py:405: error: Argument 2 to "get_s5_initialisation_times" of "S5Exporter" has incompatible type "Optional[int]"; expected "int"
src/exporters/s5.py:405: error: Argument 3 to "get_s5_initialisation_times" of "S5Exporter" has incompatible type "Optional[int]"; expected "int"
src/exporters/s5.py:405: error: Argument 4 to "get_s5_initialisation_times" of "S5Exporter" has incompatible type "Optional[int]"; expected "int"
src/exporters/s5.py:405: error: Argument 5 to "get_s5_initialisation_times" of "S5Exporter" has incompatible type "Optional[int]"; expected "int"
```

If I change to `Optional[int] = ...` for all the arguments in `get_s5_initialisation_times`
```
src/exporters/s5.py:220: error: Unsupported operand types for <= ("int" and "None")
src/exporters/s5.py:220: note: Left operand is of type "Optional[int]"
src/exporters/s5.py:224: error: Unsupported operand types for >= ("int" and "None")
src/exporters/s5.py:224: note: Left operand is of type "Optional[int]"
src/exporters/s5.py:229: error: Argument 1 to "range" has incompatible type "Optional[int]"; expected "int"
src/exporters/s5.py:229: error: Unsupported operand types for + ("None" and "int")
src/exporters/s5.py:229: note: Left operand is of type "Optional[int]"
src/exporters/s5.py:230: error: Argument 1 to "range" has incompatible type "Optional[int]"; expected "int"
src/exporters/s5.py:230: error: Unsupported operand types for + ("None" and "int")
src/exporters/s5.py:230: note: Left operand is of type "Optional[int]"
```

tests/exporters/test_s5.py
```diff
@@ -0,0 +1,307 @@
...
+        exp_months = [
+            "{:02d}".format(month) for month in range(min_month, max_month + 1)
+        ]
+        assert (
+            processed_selection_request["month"] == exp_months
+        ), f"\
+        Expected 'month' to be {exp_months}. Got:\
+        {processed_selection_request['month']}"
+
+    @patch('cdsapi.Client')
```
tommylees112: The `patch` here is causing some funny errors

```
====================================================================== FAILURES ======================================================================
_______________________________________________________ TestS5Exporter.test_expected_filepath ________________________________________________________

self = <tests.exporters.test_s5.TestS5Exporter object at 0x7f1688d5db00>, tmp_path = <MagicMock name='Client' id='139734761829656'>

    @patch('cdsapi.Client')
    def test_expected_filepath(self, tmp_path):
        granularity = "monthly"
        pressure_level = False

        s5 = S5Exporter(
            data_folder=tmp_path, granularity=granularity, pressure_level=pressure_level
        )

        expected_filepath = (
            tmp_path / "raw/seasonal-monthly-single-levels/\
            total_precipitation/2017_2018/M01_12.grib"
        ).as_posix()
        expected_filepath = expected_filepath.replace(' ', '')

        variable = "total_precipitation"
        max_leadtime = 5
        min_year = 2017
        max_year = 2018
        min_month = 1
        max_month = 12

        processed_selection_request = s5.create_selection_request(
            variable=variable,
            max_leadtime=max_leadtime,
            min_year=min_year,
            max_year=max_year,
            min_month=min_month,
            max_month=max_month,
        )

        filepath = s5.make_filename(
            dataset=s5.dataset, selection_request=processed_selection_request
        )

>       assert (
            expected_filepath == filepath.as_posix()
        ), f"\
            Expected: {expected_filepath}. Got: {filepath.as_posix()}"
E       AssertionError:             Expected: <MagicMock name='Client.__truediv__().as_posix().replace()' id='139734761441040'>. Got: <MagicMock name='Client.__truediv__().__truediv__().__truediv__().__truediv__().__truediv__().as_posix()' id='139734779376024'>
E       assert <MagicMock na...734761441040'> == <MagicMock nam...734779376024'>
E         -<MagicMock name='Client.__truediv__().as_posix().replace()' id='139734761441040'>
E         +<MagicMock name='Client.__truediv__().__truediv__().__truediv__().__truediv__().__truediv__().as_posix()' id='139734779376024'>
E         Full diff:
E         - <MagicMock name='Client.__truediv__().as_posix().replace()' id='139734761441040'>
E         + <MagicMock name='Client.__truediv__().__truediv__().__truediv__().__truediv__().__truediv__().as_posix()' id='139734779376024'>

tests/exporters/test_s5.py:244: AssertionError
```

src/exporters/s5.py
gabrieltseng: So the trouble was that the default gets overridden by the None argument being passed to `get_s5_initialization_times`. The solution was to not pass the argument it if was None.

src/exporters/s5.py
gabrieltseng: Fixed by adding type hints

src/exporters/cds.py
```diff
@@ -241,7 +249,8 @@ def export(self,
                granularity: str = 'hourly',
                show_api_request: bool = True,
                selection_request: Optional[Dict] = None,
-               break_up: bool = True) -> List[Path]:
+               break_up: bool = True,
+               N_parallel_requests: int = 3) -> List[Path]:
```
gabrieltseng: Nit: argument names should be lowercase. Can we rename this to `n_parallel_requests`?

src/exporters/s5.py
```diff
@@ -0,0 +1,492 @@
...
+        max_month: int = 12,
+        max_leadtime: Optional[int] = None,
+        pressure_levels: Optional[List[int]] = None,
+        selection_request: Optional[Dict] = None,
+        N_parallel_requests: int = 3,
+        show_api_request: bool = True,
+    ):
+        """
+        Arguments
+        --------
```
gabrieltseng: Nit: This docstring is missing some of the parameters

src/exporters/s5.py
```diff
@@ -0,0 +1,492 @@
...
+        self,
+        variable: str,
+        min_year: int = 2017,
+        max_year: int = 2018,
+        min_month: int = 1,
+        max_month: int = 12,
+        max_leadtime: Optional[int] = None,
+        pressure_levels: Optional[List[int]] = None,
+        selection_request: Optional[Dict] = None,
+        N_parallel_requests: int = 3,
```
gabrieltseng: Nit: same as above re: capitalized argument name

src/exporters/s5.py
```diff
@@ -0,0 +1,492 @@
...
+            product_type = "monthly_mean"
+
+        assert (
+            product_type in valid_product_types
+        ), f"Invalid `product_type`: \
+            {product_type}. Must be one of: {valid_product_types}"
+
+        return product_type
+
+    # @staticmethod
```
gabrieltseng: Can this be deleted?

src/exporters/s5.py
```diff
@@ -0,0 +1,492 @@
...
+        # get the leadtime information
+        leadtimes_dict = self.get_s5_leadtimes(max_leadtime)
+        for key, val in leadtimes_dict.items():
+            processed_selection_request[key] = val
+
+        # by default, we investigate Kenya
+        kenya_region = get_kenya()
+        processed_selection_request["area"] = self.create_area(kenya_region)
+
+        # TODO: do we want this level of flexibility for the user? why add complexity?
```
gabrieltseng: If we are commenting this out, can we remove the input argument from the function? Otherwise, it will be confusing to have a user-defined `selection_request` which we just ignore.

I'm happy for this to be removed from the `ERA5Exporter` too if you think that makes more sense

src/exporters/all_valid_s5.py
```diff
@@ -0,0 +1,216 @@
+"""
```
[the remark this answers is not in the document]
gabrieltseng: Thanks for doing this though. Much more robust

src/exporters/s5.py
```diff
@@ -0,0 +1,492 @@
...
+            N_parallel_requests = 1
+
+        # max_leadtime defaults
+        if max_leadtime is None:
+            # set the max_leadtime to 3 months as default
+            max_leadtime = 90 if (self.granularity == "hourly") else 3
+
+        if pressure_levels is None:
+            # set the pressure_levels to ['200', '500', '925'] as default
+            pressure_levels = [200, 500, 925] if (self.pressure_level) else None
```
gabrieltseng: Nit: parentheses around `self.pressure_level` unnecessary

src/exporters/s5.py
```diff
@@ -0,0 +1,492 @@
...
+
+        processed_selection_request = self.create_selection_request(
+            variable=variable,
+            max_leadtime=max_leadtime,
+            min_year=min_year,
```
gabrieltseng: The values for `min_year` differ between different functions. For the export function, it is 2018, for `get_s5_initialization_times` they are 1993, for `create_selection_request` it is 1993. Can this be normalized?

src/exporters/s5.py
```diff
@@ -0,0 +1,492 @@
...
+                processed_selection_request.update(cast(Dict, pressure_levels_dict))
+
+        if N_parallel_requests > 1:  # Run in parallel
+            # p = multiprocessing.Pool(int(N_parallel_requests))
+            p = pool(int(N_parallel_requests))  # pathos seems to pickle classes
+
+        # SPLIT THE API CALLS INTO MONTHS (speed up downloads)
+        output_paths = []
+        for year, month in itertools.product(
+            processed_selection_request["year"], processed_selection_request["month"]
```
gabrieltseng: Nice this is clever. Should it be optional though, using a `break_up` input argument?

src/exporters/s5.py
```diff
@@ -0,0 +1,492 @@
...
+            Must be one of:{self.dataset_reference['pressure_level']}"
+
+            if pressure_levels is not None:  # (s0rry) for mypy
+                return {"pressure_level": [str(pl) for pl in pressure_levels]}
+            else:
+                return None
+        else:
+            return None
+
+    def get_valid_variables(self) -> List:
```
gabrieltseng: This can be a property:
```python
    @property
    def valid_variables(self) -> List:
        return self.dataset_reference["variable"]
```
This way, other functions can call it as
```python
valid_variables = self.valid_variables
```

src/exporters/s5.py
```diff
@@ -0,0 +1,492 @@
...
+                else "seasonal-monthly-single-levels"
+            )
+        elif granularity == "hourly":
+            return (
+                "seasonal-original-pressure-levels"
+                if pressure_level
+                else "seasonal-original-single-levels"
+            )
+        else:
+            assert False, f"No dataset can be created with \
```
gabrieltseng: It seems like `granularity` (which is defined at initialization) can only be `monthly` or `hourly`. Would it make sense to check that the string being passed at initialization is one of the two, instead of here?
This way, the check only happens once, straight away.

tests/exporters/test_s5.py
```diff
@@ -0,0 +1,308 @@
...
+            max_year=max_year,
+            min_month=min_month,
+            max_month=max_month,
+            max_leadtime=max_leadtime,
+            show_api_request=show_api_request,
+            N_parallel_requests=N_parallel_requests,
+        )
+
+        (
+            tmp_path / "raw/seasonal-monthly-single-levels\
```
gabrieltseng: What's happening here? Is an `assert` statement missing?

tests/exporters/test_s5.py
```diff
@@ -0,0 +1,308 @@
+# from pathlib import Path
```
gabrieltseng: Nit: This import can be removed

tommylees112: thanks

tests/exporters/test_s5.py
tommylees112: @gabrieltseng i don't know how to test this because i don't actually want to download the file (because it will take like 24hrs + ... ) but i want to mock that behaviour

src/exporters/s5.py
tommylees112: great idea

src/exporters/s5.py
tommylees112: cool!

src/exporters/s5.py
tommylees112: haha this is your code 👍 (i agree its v.clever)

Do we definitely want this to be an option? This is one of teh things i think we probably want to assume for the user?

Although saying that you might be right because i don't think its the size of the data download limiting speedds. i think i'm somehow at the bottom of the food chain

tommylees112: okay added it in as an option

tommylees112: might be interesting to test running this in parallel outside of the class for example

src/exporters/s5.py
tommylees112: done!

src/exporters/all_valid_s5.py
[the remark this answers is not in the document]
tommylees112: nice idea trying to implement now

src/exporters/s5.py
tommylees112: thanks

src/exporters/cds.py
tommylees112: definitley

src/exporters/s5.py
tommylees112: if you're happy for me to remove that functionality from the user?

tommylees112: sorry just read below

src/exporters/s5.py
tommylees112: i'm not sure - for this one i think it does (with the current structure like the `all_valid_s5.py` where i know we have captured every use case. the only issue is that some users might want to use our scaffolding (eg the parallelisation) but withtheir preprepared api calls. what do you think?

gabrieltseng: Yea, when I had the selection request argument my main intention was if the user wanted to define their own times, but I think you have handled that by adding the `min/max_year/month` arguments, so its no longer really necessary

gabrieltseng: I can delete it

tests/exporters/test_s5.py
gabrieltseng: Added an assert statement to make sure the cds api is called

Issue #8, opened 2019-05-10
Comparing forecast to Parsimonious model


tommylees112:
**Persistence** - For june prediction simply take the values from May

**Mean** - for that month (or timestep) take the mean across all years

PR #28, opened 2019-05-31
Feature engineering

gabrieltseng:


src/engineer/engineer.py
```diff
@@ -0,0 +1,146 @@
...
+    @staticmethod
+    def _train_test_split_single_year(ds: xr.Dataset,
+                                      year: int,
+                                      target_variable: str,
+                                      target_month: int
+                                      ) -> Tuple[xr.Dataset, xr.Dataset]:
+
+        print(f'Generating test data for year: {year}, target month: {target_month}')
+        max_date = date(year, target_month, calendar.monthrange(year, target_month)[-1])
+        min_date = max_date - relativedelta(years=1)
```
tommylees112: love the `relativedelta` usage

tests/engineer/test_engineer.py
```diff
@@ -0,0 +1,109 @@
...
+            filename = interim_folder / f'{var}_preprocessed/hello.nc'
+            data.to_netcdf(filename)
+
+            expected_output.append(filename)
+            expected_vars.append(var)
+
+            # this file should not
+            (interim_folder / f'{var}_preprocessed/boo').touch()
+
+        # none of this should be captured
```
tommylees112: so what is this testing?

src/engineer/engineer.py
```diff
@@ -0,0 +1,146 @@
...
+        assert self.interim_folder.exists(), \
+            f'{data_folder / "interim"} does not exist. Has the preprocesser been run?'
+
+        self.output_folder = data_folder / 'features'
+        if not self.output_folder.exists():
+            self.output_folder.mkdir()
+
+    def engineer(self, test_year: Union[int, List[int]],
+                 target_variable: str = 'VHI',
+                 target_month: int = 6):
```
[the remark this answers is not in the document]
gabrieltseng: Hmm good point. I'll remove the target month argument, and we'll have a year of test data to predict.

tests/engineer/test_engineer.py
gabrieltseng: This function just sets up stuff (since it's name doesn't start with `test`, no tests are run on it). It's used by all the other tests.

src/engineer/engineer.py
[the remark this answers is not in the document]
gabrieltseng: I've also split the test data so there is a `test_x`, which the model gets as input, and a `test_y`, which it tries to predict.

gabrieltseng:
Going to build the first model off this branch, since its likely that how we structure our models will affect how the feature engineering should look like

src/engineer/engineer.py
```diff
@@ -0,0 +1,181 @@
...
+            Data to be used for testing. No data earlier than the earliest test year
+            will be used for training.
+            If a list is passed, a file for each year will be saved.
+        target_variable: str = 'VHI'
+            The variable to be predicted. Only this variable will be saved in the test
+            netcdf files
+        """
+        data = self._make_dataset()
+
+        if type(test_year) is int:
```
tommylees112: if test_year is of type `int` then cast test_year to `int`?

gabrieltseng: So if it is an int, I turn it into a list.

The `cast` bit it just to tell mypy that if we're here, we're sure `year` is an `int`, not a `Union[List[int], int]`.

gabrieltseng:
Implements the persistence model described in #8

src/engineer/engineer.py
```diff
@@ -0,0 +1,229 @@
...
+            return None, output_min_date
+
+        if expected_length is not None:
+            if sum(x) != expected_length:
+                print(f'Wrong number of test x values! Got {sum(x)} Returning None')
+
+                return None, output_min_date
+
+        x_dataset = ds.isel(time=x)
+        y_dataset = ds.isel(time=y)[target_variable].to_dataset()
```
tommylees112: We should give the to_dataset() a name in case the `DataArray` is unnamed ( it will throw an error if the `DataArray` is unnamed).

http://xarray.pydata.org/en/stable/generated/xarray.DataArray.to_dataset.html

```
ds.isel(time=y)[target_variable].to_dataset(name=target_variable)
```

Although it might not be necessary because we've already selected that `target_variable`.

IT's just an error i have come across before

src/engineer/engineer.py
```diff
@@ -0,0 +1,229 @@
...
+            Data to be used for testing. No data earlier than the earliest test year
+            will be used for training.
+            If a list is passed, a file for each year will be saved.
+        target_variable: str = 'VHI'
+            The variable to be predicted. Only this variable will be saved in the test
+            netcdf files
+        pred_months: int = 11
+            The amount of months of data to feed as input to the model for it to make its
+            prediction
+        expected_length: Optional[int] = 11
```
tommylees112: could we make the `expected_length` docstring a little clearer?

I don't really understand how it differs from `pred_months`? Why would we assign `pred_months=11` but then `expected_length=4` for example?

src/engineer/engineer.py
```diff
@@ -0,0 +1,229 @@
...
+
+        save_folder = self.output_folder / dataset_type
+        save_folder.mkdir(exist_ok=True)
+
+        for year_key, val in ds.items():
+
+            for month_key, test_dict in val.items():
+
+                output_location = save_folder / f'{year_key}_{month_key}'
+                output_location.mkdir()
```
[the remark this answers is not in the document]
gabrieltseng: This works! Thanks for making the change :)

src/engineer/engineer.py
gabrieltseng: `expected_length` depends on the time granularity of the data. So here, because its monthly, `expected_length == pred_months`. But if it was weekly, it would be `expected_length == 4 * pred_months`

gabrieltseng: I'll update the docstring too

src/engineer/engineer.py
```diff
@@ -0,0 +1,229 @@
...
+        for idx, file in enumerate(self._get_preprocessed_files()):
+            print(f'Processing {file}')
+            datasets.append(xr.open_dataset(file))
+
+            if idx == 0:
+                for dim in dims:
+                    coords[dim] = datasets[idx][dim].values
+            else:
+                for dim in dims:
+                    assert (datasets[idx][dim].values == coords[dim]).all(), \
```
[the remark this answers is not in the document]
gabrieltseng: Weird... updated to use `np.array_equals` to compare the two arrays.

src/engineer/engineer.py
gabrieltseng: Updated!

src/engineer/engineer.py
tommylees112: awesome thanks mate

PR #34, opened 2019-06-05
Add preprocessor for Planet OS

gabrieltseng:


tommylees112:
Added time information (year, month) to the `print()` statement in the `_rotate_and_filter` because it's useful to know when you have all of the data there (2008-2019) as it takes quite a while (~300GB)

tommylees112:
problems working in parallel because reading and writing from the same file - added a random number to the end of the regridding filename which seems to have worked

PR #35, opened 2019-06-06
GLEAM exporter

gabrieltseng:
Adds a GLEAM exporter


scripts/export.py
```diff
@@ -57,7 +59,25 @@ def export_era5POS():
...
 
 
+def export_gleam():
+    # if the working directory is alread ml_drought don't need ../data
+    if Path('.').absolute().as_posix().split('/')[-1] == 'ml_drought':
+        data_path = Path('data')
+    else:
+        data_path = Path('../data')
+
+    host = 'Get the login details'
```
gabrieltseng: @tommylees112 , it seems like the login details are uniform (the ones you got were the same as what I got); we could just add them to the GLEAM exporter class.

src/exporters/gleam.py
```diff
@@ -0,0 +1,93 @@
...
+
+        return output_datasets
+
+    def sftppath_to_localpath(self, sftppath: str) -> Tuple[Path, str]:
+
+        stem = sftppath[len(self.base_sftp_path):]
+        filename = sftppath.split('/')[-1]
+        return self.gleam_folder / stem[:-len(filename)].strip('/'), filename
+
+    def export(self,
```
tommylees112: can we have a parallel exporter too?

gabrieltseng: Since there are not many files to download, do you think its necessary? (i.e. for `monthly` or `yearly` granularities, there is only one file per variable, and for `daily` there is only one file per variable per year)

tommylees112: happy with that

scripts/export.py
tommylees112: yeah i think we should - can you make that change? or add to the script example?

gabrieltseng: I'll make the change!

gabrieltseng: Updated

PR #40, opened 2019-06-10
Preprocess/subsets

tommylees112:
using different subsets of the data

### DIFF
--- a/src/preprocess/base.py
+++ b/src/preprocess/base.py
@@ -5,8 +5,8 @@ import numpy as np
 
 from typing import List, Optional
 
-from ..utils import Region, get_kenya
-
+from ..utils import Region, get_kenya, region_lookup
+from .utils import select_bounding_box
 
 __all__ = ['BasePreProcessor', 'Region', 'get_kenya']
 
@@ -145,7 +145,16 @@ class BasePreProcessor:
         else:
             return resampler.nearest()
 
-    def merge_files(self, subset_kenya: bool = True,
+    @staticmethod
+    def chop_roi(ds: xr.Dataset,
+                 subset_str: str = 'kenya',) -> xr.Dataset:
+        region = region_lookup[subset_str] if subset_str is not None else None
+        if region is not None:
+            ds = select_bounding_box(ds, region)
+
+        return ds
+
+    def merge_files(self, subset_str: str = 'kenya',
                     resample_time: Optional[str] = 'M',
                     upsampling: bool = False) -> None:
 
@@ -154,6 +163,6 @@ class BasePreProcessor:
         if resample_time is not None:
             ds = self.resample_time(ds, resample_time, upsampling)
 
-        out = self.out_dir / f'{self.dataset}{"_kenya" if subset_kenya else ""}.nc'
+        out = self.out_dir / f'{self.dataset}_{subset_str if subset_str is not None else ""}.nc'
         ds.to_netcdf(out)
         print(f"\n**** {out} Created! ****\n")
--- a/src/preprocess/base.py
+++ b/src/preprocess/base.py
@@ -5,10 +5,10 @@ import numpy as np
 
 from typing import List, Optional
 
-from ..utils import Region, get_kenya, region_lookup
+from ..utils import Region, region_lookup
 from .utils import select_bounding_box
 
-__all__ = ['BasePreProcessor', 'Region', 'get_kenya']
+__all__ = ['BasePreProcessor', 'Region']
 
 
 class BasePreProcessor:
@@ -168,6 +168,6 @@ class BasePreProcessor:
         if resample_time is not None:
             ds = self.resample_time(ds, resample_time, upsampling)
 
-        out = self.out_dir / f'{self.dataset}_{subset_str if subset_str is not None else ""}.nc'
+        out = self.out_dir / f'{self.dataset}{"_" + subset_str if subset_str is not None else ""}.nc'
         ds.to_netcdf(out)
         print(f"\n**** {out} Created! ****\n")
--- a/src/preprocess/base.py
+++ b/src/preprocess/base.py
@@ -147,7 +147,7 @@ class BasePreProcessor:
 
     @staticmethod
     def chop_roi(ds: xr.Dataset,
-                 subset_str: str = 'kenya',
+                 subset_str: Optional[str] = 'kenya',
                  inverse_lat: bool = False) -> xr.Dataset:
         """ lookup the region information from the dictionary in
         `src.utils.region_lookup` and subset the `ds` object based on that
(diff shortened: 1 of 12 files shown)


tommylees112:
NOTE TO SELF: before merging make sure that we fix the 'default' experiments in the `pipeline_config` dir

Now passing a `string` instead of `boolean` for the` subset_kenya` => `subset_str` argument

src/preprocess/base.py
```diff
@@ -5,8 +5,8 @@
 
 from typing import List, Optional
 
-from ..utils import Region, get_kenya
-
+from ..utils import Region, get_kenya, region_lookup
```
gabrieltseng: Since we are using the `region_lookup`, is `get_kenya` a necessary import here?

Equally, is it necessary in `__all__`?

src/preprocess/vhi.py
```diff
@@ -19,14 +19,11 @@
 
 from xarray import Dataset, DataArray
 
-from .base import (BasePreProcessor, get_kenya)
+from .base import (BasePreProcessor,)
```
gabrieltseng: nit: the brackets around the import aren't really necessary

src/preprocess/chirps.py
```diff
@@ -74,7 +72,7 @@ def create_filename(netcdf_filepath: str,
             new_filename = f"{filename_stem}.nc"
         return new_filename
 
-    def preprocess(self, subset_kenya: bool = True,
+    def preprocess(self, subset_str: str = 'kenya',
```
gabrieltseng: Should we make this `Optional`?

src/preprocess/chirps.py
```diff
@@ -19,7 +18,7 @@ class CHIRPSPreprocesser(BasePreProcessor):
     dataset = 'chirps'
 
     def _preprocess_single(self, netcdf_filepath: Path,
-                           subset_kenya: bool = True,
+                           subset_str: str = 'kenya',
```
gabrieltseng: Should be `Optional`

src/preprocess/base.py
```diff
@@ -145,7 +145,21 @@ def resample_time(ds: xr.Dataset,
...
+        `src.utils.region_lookup` and subset the `ds` object based on that
+        region.
+        """
+        region = region_lookup[subset_str] if subset_str is not None else None
+        if region is not None:
+            ds = select_bounding_box(ds, region, inverse_lat=inverse_lat)
+
+        return ds
+
+    def merge_files(self, subset_str: str = 'kenya',
```
gabrieltseng: Should be `Optional`

src/preprocess/gleam.py
```diff
@@ -23,7 +22,7 @@ def _swap_dims_and_filter(ds: xr.Dataset) -> xr.Dataset:
         return ds.transpose(*expected_dims)
 
     def _preprocess_single(self, netcdf_filepath: Path,
-                           subset_kenya: bool = True,
+                           subset_str: str = 'kenya',
```
gabrieltseng: Should be `Optional`

src/preprocess/planetOS.py
```diff
@@ -68,7 +67,7 @@ def create_filename(netcdf_filepath: Path,
         return f'{stem}.nc'
 
     def _preprocess_single(self, netcdf_filepath: Path,
-                           subset_kenya: bool = True,
+                           subset_str: str = 'kenya',
```
gabrieltseng: Should be `Optional`

src/preprocess/planetOS.py
```diff
@@ -77,24 +76,22 @@ def _preprocess_single(self, netcdf_filepath: Path,
...
-            subset_name='kenya' if subset_kenya else None
+            subset_name=subset_str if subset_str is not None else None
         )
         print(f'Saving to {self.interim}/{filename}')
         ds.to_netcdf(self.interim / filename)
 
         print(f'Done for ERA5 Planet OS {netcdf_filepath.name}')
 
-    def preprocess(self, subset_kenya: bool = True,
+    def preprocess(self, subset_str: str = 'kenya',
```
gabrieltseng: Should be `Optional`

src/preprocess/vhi.py
```diff
@@ -85,7 +81,7 @@ def _preprocess(self,
 
     def _preprocess_wrapper(self,
                             netcdf_filepath: str,
-                            subset_kenya: bool = True,
+                            subset_str: str = 'kenya',
```
gabrieltseng: Should be `Optional`

src/preprocess/vhi.py
```diff
@@ -101,14 +97,14 @@ def _preprocess_wrapper(self,
...
-                netcdf_filepath, self.interim.as_posix(), subset_kenya, regrid
+                netcdf_filepath, self.interim.as_posix(), subset_str, regrid
             )
         except Exception as e:
             print(f"### FAILED: {netcdf_filepath}")
             return e, netcdf_filepath
 
     def preprocess(self,
-                   subset_kenya: bool = True,
+                   subset_str: str = 'kenya',
```
gabrieltseng: Should be `Optional`

tests/preprocess/test_base.py
```diff
@@ -65,3 +65,18 @@ def test_load_regridder(self, tmp_path):
...
+        original_ds, _, _ = _make_dataset(size_original)
+
+        original_shape = original_ds.VHI.shape
+
+        processor = BasePreProcessor(tmp_path)
+        subset_str = 'east_africa'
+        new_ds = processor.chop_roi(ds=original_ds, subset_str=subset_str)
+        output_shape = new_ds.VHI.shape
+
+        assert original_shape != output_shape, f"The chop_roi should lead to\
```
gabrieltseng: Can we also add an assert here to check the latitudes and longitudes are as expected?

gabrieltseng: (Especially because all the preprocessor tests don't check the chopping happens correctly, only that some output is saved)

src/preprocess/base.py
tommylees112: nope ! good spot

tests/preprocess/test_base.py
tommylees112: very true. thanks Gabi!

tommylees112:
```
        assert (
            (new_ds.lat.values.min() >= -11) & (new_ds.lat.values.max() <= 23)
        ), f"Expected latitude to be in the range -11 : 23. Currently:\
        {new_ds.lat.values.min()} : {new_ds.lat.values.max()}"

        assert (
            (new_ds.lon.values.min() >= 21) & (new_ds.lon.values.max() <= 51.8)
        ), f"Expected longitude to be in the range 21 : 51.8. Currently:\
        {new_ds.lon.values.min()} : {new_ds.lon.values.max()}"
```

PR #41, opened 2019-06-10
parallel cds exporter - rename `n_parallel_requests` arg

tommylees112:
MINIMAL CHANGE (just passing the tests and things (because of local difficulties)


tommylees112:
Just changing the case from camel/snake case to snake case for the `n_parallel_requests` arg

PR #42, opened 2019-06-10
Add option to export vhi files serially

gabrieltseng:
As described. Also, an argument to define how many processes are created.


gabrieltseng:
Hmm, yea...
It will require some serious mocking though, and the functionality is pretty well tested. We can add that as a todo?

gabrieltseng:
https://trello.com/c/wOvBLI0D/30-unit-testing-end-to-end-functionality-of-the-vhi-exporter

PR #43, opened 2019-06-11
Modelling/nowcasting expt

tommylees112:


tommylees112:
Hey @gabrieltseng can you help me with the final test 

in `tests/engineer/test_one_month_forecast.py`
```    def test_engineer(self, tmp_path):  ```

I seem to be getting an error with the creation of the files. if you can let me know how it works and how to fix that error i can have a go at reimplementing the same tests for the `nowcast` engineer too

File: src/engineer/__init__.py [hunk 1]
```diff
@@ -1,3 +1,4 @@
 from .engineer import Engineer
```
gabrieltseng: No need to import the Engineer, since it is only used when its inherited by the other classes

src/engineer/nowcast.py
```diff
@@ -0,0 +1,124 @@
...
+        We overcome this by creating an extra timestep with all nan values
+        in the `x_dataset`. This way the `x_dataset` contains the `y_dataset`
+        timestep but the `target_variable` is an array of all `np.nan` for that
+        target timestep. This prevents model leakage.
+        """
+        if expected_length is not None:
+            warnings.warn("Expected length must be None\
+                          for the `nowcast` experiment. Forcing to\
+                          None")
+            expected_length = None
```
gabrieltseng: As we discussed, the expected length check got triggered quite often. Is there a way to consider it anyway?

One option would be to have it for the `non_target_dataset`, and then `expected_length + 1` for the `target_dataset`

src/engineer/nowcast.py
```diff
@@ -0,0 +1,124 @@
...
+        nan_target_variable = self.make_nan_dataset(y_dataset)
+        x_target_dataset = (
+            ds[target_variable].isel(time=x_target).to_dataset(name=target_variable)
+        )
+        x_target_dataset = x_target_dataset.merge(nan_target_variable)
+
+        # merge the x_non_target_dataset + x_target_dataset -> x_dataset
+        x_dataset = x_non_target_dataset.merge(x_target_dataset)
+
+        if x_dataset.time.size != pred_months + 1:
```
gabrieltseng: This assumes the preprocessors were run with monthly granularity. We probably want to compare this to `expected_length`, not `pred_months`

tests/engineer/test_engineer.py
```diff
@@ -1,7 +1,7 @@
-import pytest
-import pickle
-import numpy as np
-import xarray as xr
+# import pytest
```
gabrieltseng: Can we remove these unused imports?

tests/engineer/test_nowcast.py
```diff
@@ -0,0 +1,38 @@
+import pytest
+# import pickle
```
gabrieltseng: Can we remove these unused imports?

tests/engineer/test_nowcast.py
```diff
@@ -0,0 +1,38 @@
...
+import numpy as np
+# import xarray as xr
+
+from src.engineer import NowcastEngineer
+
+from ..utils import _make_dataset
+from .test_engineer import TestEngineer
+
+
+class TestNowcastEngineer(TestEngineer):
```
gabrieltseng: Since the `stratify_xy` functions differ between the engineers, could we add tests to make sure they work?

gabrieltseng: (Noticing now you have already mentioned this)

tests/engineer/test_engineer.py
tommylees112: thanks

src/engineer/nowcast.py
tommylees112: okay so i removed this line and added the following below:

```
        if expected_length is not None:
            # filter for missing values in timesteps!
            if sum(x_target_dataset) != expected_length:
                print(f"Wrong number of x values! Got {sum(x)} Returning None")

                return None, cast(date, max_train_date)

            if sum(x_non_target_dataset) != expected_length + 1:
                print(f"Wrong number of x values! Got {sum(x)} Returning None")

                return None, cast(date, max_train_date)

```

src/engineer/nowcast.py
tommylees112: so now the choice of `expected_length = None` as default looks off that's my bad!

does `pred_months` work with non-month timesteps because it's filtering with time instead of number of steps?

tommylees112: i now understand why its so important

tests/engineer/test_nowcast.py
tommylees112: todo thanks!

tommylees112: done thanks mate

src/engineer/nowcast.py
tommylees112: oaky added a warning if the value is `None` too

gabrieltseng: Thanks Tommy!

src/engineer/one_month_forecast.py
```diff
@@ -0,0 +1,77 @@
...
+        super().__init__(self.name, data_folder)
+
+    @staticmethod
+    def stratify_xy(
+        ds: xr.Dataset,
+        year: int,
+        target_variable: str,
+        target_month: int,
+        pred_months: int,
+        expected_length: Optional[int],
```
gabrieltseng: Can the default values be put back here too?

gabrieltseng:
Just the last change with the default values in the one month forecasting engineer; then it looks good to me!

src/engineer/one_month_forecast.py
tommylees112: absolutely!

PR #49, opened 2019-06-30
Add nowcasting option to the pipeline

tommylees112:


src/models/parsimonious.py
```diff
@@ -21,10 +21,15 @@ def train(self) -> None:
     def save_model(self) -> None:
         print('Move on! Nothing to save here!')
 
-    def predict(self) -> Tuple[Dict[str, Dict[str, np.ndarray]], Dict[str, np.ndarray]]:
-
-        test_arrays_loader = DataLoader(data_path=self.data_path, batch_file_size=self.batch_size,
-                                        shuffle_data=False, mode='test', normalize=False)
+    def predict(
+        self,
+        experiment: str = 'one_month_forecast'
```
gabrieltseng: Howcome this is passed as an argument to `predict`? Would it make more sense as a class attribute?

Could it even be hardcoded given the nature of this "model"?

src/models/regression.py
```diff
@@ -14,12 +14,14 @@ class LinearRegression(ModelBase):
     model_name = 'linear_regression'
 
     def train(self, num_epochs: int = 1,
+              experiment: str = 'one_month_forecast',
```
gabrieltseng: This should be a class attribute, since this information will also be needed by the `predict` function.

(This also needs to be added to the predict function)

src/models/linear_network.py
```diff
@@ -19,8 +19,10 @@ class LinearNetwork(ModelBase):
     def __init__(self, layer_sizes: Union[int, List[int]],
                  dropout: float = 0.25,
                  data_folder: Path = Path('data'),
+                 experiment: str = 'one_month_forecast',
```
gabrieltseng: This should be a class attribute (i.e. added to the `__init__` function) since it will also be used by the predict function.

It should also be added to the predict function

scripts/models.py
```diff
@@ -29,7 +29,7 @@ def regression():
     else:
         data_path = Path('../data')
 
-    predictor = LinearRegression(data_path)
+    predictor = LinearRegression(data_path, experiment='one_month_forecast')
```
gabrieltseng: None of these are in the `__init__` methods in the actual models, so this script doesn't work

gabrieltseng: (This is also causing the tests to fail: `TypeError: __init__() got an unexpected keyword argument 'experiment'`)

src/models/parsimonious.py
tommylees112: thanks for finding these mate!

tommylees112: great spot

tommylees112: Re: hardcoded - it could be used as a baseline for any model? So even for the `nowcast` experiment persistence would still be a reasonable baseline to compare against?

src/models/linear_network.py
tommylees112: I added to the `base` model because every model will now have to have an experiment string to define where it is reading the data from

scripts/models.py
tommylees112: this is fixed

tommylees112:
Need to fix the creation of the `normalizing_dict` because it responds to the `nan` VHI in the target year and therefore leaves the VHI `mean` and `std` keys out of the `normalizing_dict` object:

```
with open(data_dir/'features'/ 'nowcast'/'normalizing_dict.pkl', 'rb') as f:
    n = pickle.load(f)

defaultdict(dict,
            {'precip': {'mean': array([26.438026, 26.410048, 26.522446, 26.808918], dtype=float32),
              'std': array([24.444538, 24.392359, 24.414604, 24.57235 ], dtype=float32)},
             'E': {'mean': array([22.3748105 , 22.35253429, 22.40610263, 22.47338553]),
              'std': array([16.16524859, 16.18476273, 16.17894641, 16.13133825])},
             'SMroot': {'mean': array([0.10808951, 0.10799525, 0.10798577, 0.10805881]),
              'std': array([0.03803128, 0.0380079 , 0.03799228, 0.03795943])},
             'SMsurf': {'mean': array([0.11100077, 0.11091969, 0.11100009, 0.11121675]),
              'std': array([0.04285916, 0.04283779, 0.0427892 , 0.04272659])}})
```

tommylees112:
the `notnan_indices` [here: src/models/data.py#L190](https://github.com/esowc/ml_drought/blob/25aa060a5b65158f8b88cb225e3214800e428faa/src/models/data.py#L190) asks for BOTH `x==np.nan` and `y==np.nan`

```python
notnan_indices = np.where((x_nans_summed == 0) & (y_nans_summed == 0))[0]
```

But in the nowcast experiment only the `x` variable has `nan`. So it remains. 

I can't decide if this is good or bad? But it means that the `nan` `x_vals` remain in the training / test data and causes problems downstream

tommylees112:
> Need to fix the creation of the `normalizing_dict` because it responds to the `nan` VHI in the target year and therefore leaves the VHI `mean` and `std` keys out of the `normalizing_dict` object:
> 
> ```
> with open(data_dir/'features'/ 'nowcast'/'normalizing_dict.pkl', 'rb') as f:
>     n = pickle.load(f)
> 
> defaultdict(dict,
>             {'precip': {'mean': array([26.438026, 26.410048, 26.522446, 26.808918], dtype=float32),
>               'std': array([24.444538, 24.392359, 24.414604, 24.57235 ], dtype=float32)},
>              'E': {'mean': array([22.3748105 , 22.35253429, 22.40610263, 22.47338553]),
>               'std': array([16.16524859, 16.18476273, 16.17894641, 16.13133825])},
>              'SMroot': {'mean': array([0.10808951, 0.10799525, 0.10798577, 0.10805881]),
>               'std': array([0.03803128, 0.0380079 , 0.03799228, 0.03795943])},
>              'SMsurf': {'mean': array([0.11100077, 0.11091969, 0.11100009, 0.11121675]),
>               'std': array([0.04285916, 0.04283779, 0.0427892 , 0.04272659])}})
> ```

We don't need to change the `normalizing_dict` if we can get the target variable from the `ds` and then create two normalizing arrays 

`self.historical_normalizing_array` & `self.current_normalizing_array`

tommylees112:
Okay done (an initial attempt) at getting the `DataLoader` to work with the `nowcast` experiment but might need some help working with the `_train_iter` etc. because we now have two `X` objects - `train_data.historical` and `train_data.current`

tommylees112:
I feel like one danger is we are losing flexibility of our approach with these complexities in the `DataLoader` - what's your feeling /experience here? Is this normal for a pipeline that we want to be relatively extensible?

tommylees112:
[src/models/data #L259](https://github.com/esowc/ml_drought/blob/25aa060a5b65158f8b88cb225e3214800e428faa/src/models/data.py#L259)

Working in the `_TestIter` trying to return the `historical` and `current` arrays from the `ModelArrays`
```python
>>> arrays.x.current.shape
(35134, 1, 4)
>>> arrays.x.historical.shape
(35134, 3, 5)
```

How do we combine / use both of these `X` matrices into the training / test data with different number of `variables` (`axis=2`) ?

tommylees112:
Perhaps do we need different training regimes for the `nowcast` experiment? Or is that overkill? Is there any way to make this general?

tommylees112:
## Can we perhaps add a vector of `[1, 1, 1, 1, 1, ...]` for example?

In a linear regression setting that variable should have no variability and so should be very unimportant? We could even test the impact of the constant value added? But it shouldn't add anything right?

src/models/data.py
```diff
@@ -349,11 +349,25 @@ def __next__(self) -> Tuple[Union[np.ndarray, torch.Tensor],
 
                     cur_max_idx = min(cur_max_idx + 1, self.max_idx)
 
-                out_x.append(arrays.x)
+                out_x.append(arrays.x.historical)
+
+                if self.experiment == 'nowcast':
```
tommylees112: So this is really silly because we're basically undoing the work we did in the `ds_folder_to_np` and in the `NowcastEngineer` but it does work. Really keen to hear your thoughts @gabrieltseng

src/models/data.py
```diff
@@ -169,43 +226,109 @@ def ds_folder_to_np(self,
...
+        if self.historical_normalizing_array is not None:
+            if self.experiment == 'nowcast':
+                # normalize the historical X data (all vars including target variable)
+                train_data.historical = (
+                    train_data.historical - self.historical_normalizing_array['mean']
+                ) / (self.historical_normalizing_array['std'])
+
+                # normalize the current X data (no target variable)
+                train_data.current = (
+                    train_data.current - self.current_normalizing_array['mean']
```
gabrieltseng: The mypy errors here are because mypy still thinks `self.current_normalizing_array` could be `None` (since you don't technically check that its not in the statement above).

Mypy could be appeased if you changed the above statement to:

`if (self.historical_normalizing_array is not None) and (self.current_normalizing_array is not None):`

which should always be `True` if the current statement is

src/models/data.py
```diff
@@ -169,43 +226,109 @@ def ds_folder_to_np(self,
...
+                    current_nans.shape[0], current_nans.shape[1] * current_nans.shape[2]
+                ).sum(axis=-1)
+                notnan_indices = np.where(
+                    (
+                        historical_nans_summed == 0
+                    ) & (y_nans_summed == 0) & (
+                        current_nans_summed == 0
+                    )
+                )[0]
+                train_data.current = train_data.current[notnan_indices]
```
gabrieltseng: The mypy error here is because mypy thinks that `train_data.current` could still be None.

Casting could fix this (or, if thats a nuisance, you can ignore this line by adding the inline comment `type: ignore`)

src/models/data.py
```diff
@@ -248,13 +371,26 @@ def __next__(self) -> Tuple[Tuple[Union[np.ndarray, torch.Tensor],
                     cur_max_idx = min(cur_max_idx + 1, self.max_idx)
 
                 out_x.append(arrays.x.historical)
+
+                # @GABI we don't want this do we ? it adds a column of constants
+                if self.experiment == 'nowcast':
+                    # add a constant vector to the X train_data and append
+                    constant_vector = np.ones(arrays.x.current.shape)[:, :, :1]
+                    arrays.x.current = np.concatenate(
```
gabrieltseng: mypy error: mypy doesn't know `arrays.x.current` is not `None`. Same solution as above

src/models/data.py
```diff
@@ -169,43 +226,109 @@ def ds_folder_to_np(self,
...
         y_np = np.moveaxis(y_np, -1, 0).reshape(-1, 1)
 
-        if self.normalizing_array is not None:
-            x_np = (x_np - self.normalizing_array['mean']) / (self.normalizing_array['std'])
+        if self.experiment == 'nowcast':
+            # if nowcast then we have a TrainData.current
+            historical = x_np[:, :-1, :]  # all timesteps except the final
+            current = x_np[:, -1, :-1]  # final axis is all nan (target var)
+
+            # add 1 axis for timesteps
```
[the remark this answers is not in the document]
tommylees112: you're right we don't need to any more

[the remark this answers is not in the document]
tommylees112: it was so they could all be appended into an `final_x` array

src/models/data.py
```diff
@@ -130,34 +148,73 @@ def __init__(self, loader: DataLoader) -> None:
...
+        if self.experiment == 'nowcast':
+            historical_mean, historical_std = [], []
+            current_mean, current_std = [], []
+            for var in data_vars:
+                # historical_mean = all times up to current time
+                historical_mean.append(self.normalizing_dict[var]['mean'][:-1])
+                historical_std.append(self.normalizing_dict[var]['std'][:-1])
+
+                # current_mean = the current times for all EXCEPT for target_var
+                if var != target_var:
```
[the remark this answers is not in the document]
tommylees112: this is the right way to do it no? Because we will always have more predictor variables (Precip, Soil Moisture etc.) and only ever have one predictand (VHI)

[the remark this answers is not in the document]
tommylees112: or have i misunderstood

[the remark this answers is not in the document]
gabrieltseng: Some of the input data is not forecast data, thought.

For example, the CHIRPS data is not available into the future. This makes it a little misleading to use it in the nowcasting model, since it wouldn't be usable in a "future prediction" scenario.

I guess the question is whether the nowcasting model's "current" data should only have forecasted data (e.g. SEAS5 stuff) or whether we just want to test the model's ability to nowcast generally.

src/models/data.py
tommylees112: okay i think i fixed that

src/models/data.py
tommylees112: do you mind if i `  # type: ignore`

src/models/data.py
```diff
@@ -169,43 +226,106 @@ def ds_folder_to_np(self,
...
+                pred_months=x_months
+            )
+        else:
+            train_data = TrainData(
+                current=None,
+                historical=x_np,
+                pred_months=x_months
+            )
+
+        if self.historical_normalizing_array is not None:
```
gabrieltseng: Can this normalization take place in the if else statements above?

Makes the code a little more concise

src/models/data.py
```diff
@@ -248,13 +368,18 @@ def __next__(self) -> Tuple[Tuple[Union[np.ndarray, torch.Tensor],
                     cur_max_idx = min(cur_max_idx + 1, self.max_idx)
 
                 out_x.append(arrays.x.historical)
+
+                out_x_add.append(arrays.x.current)
```
gabrieltseng: Shouldn't this be a new list?

Otherwise, `out_x_add` will be twice as long as `out_x`

I think the zeroth dimension of the tensors / arrays should always be batch size, which they wouldn't be here

src/models/data.py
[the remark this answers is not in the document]
tommylees112: i think the current data can also have historical data (i.e. we're testing the scenario of we know the rainfall and soil moisture in this time, how can we convert these numbers into vegetation health without a physical process model?)

this will also allow us to train the model on HISTORICAL relationships between precip now -> VHI now because there might not be enough data to just use the S5 forecasts.

src/models/linear_network.py
```diff
@@ -230,6 +244,8 @@ def forward(self, x, pred_month=None):
         # flatten
         x = x.contiguous().view(x.shape[0], -1)
         if self.include_pred_month:
+            # @gabrieltseng HELP!
```
gabrieltseng: See the comment above; I think this is why we should have seperate arrays returned for the prediction month and the nowcasting stuff

src/models/data.py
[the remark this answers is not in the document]
gabrieltseng: right, that makes sense

src/models/data.py
tommylees112: same as above

src/utils.py
```diff
@@ -55,6 +55,19 @@ def minus_months(cur_year: int, cur_month: int, diff_months: int,
     return new_year, new_month, newdate
 
 
+def get_ds_mask(ds):
```
gabrieltseng: Can type hints be added here? Is the input an xarray dataset?

tests/models/test_data.py
```diff
@@ -53,37 +61,102 @@ def __init__(self):
...
+        if not to_tensor:
+            assert isinstance(y_np, np.ndarray)
+
+        for idx in range(latlons.shape[0]):
+            lat, lon = latlons[idx, 0], latlons[idx, 1]
+            for time in range(x_train_data.historical.shape[1]):
+                target = x.isel(time=time).sel(lat=lat).sel(lon=lon).VHI.values
+
+                if (not normalize) and (not to_tensor):
+                    assert target == x_train_data.historical[idx, time, 0], \
```
gabrieltseng: Can the tests also test the current arrays, since thats whats been added

src/models/data.py
gabrieltseng: yep! that works :)

src/models/data.py
tommylees112: can't it be:
```batch_size * 2 ```

so there are two vectors of batch size concatenated to each other?

Comment (gabrieltseng) [reply src/models/data.py]: The trouble is the chunking function shuffles based on the first dimension.

So when the x arrays are passed to the chunking function [here](https://github.com/esowc/ml_drought/blob/4f72a41c2c78df0c236873cddb6547dd8c5153a4/src/models/regression.py#L65), I am not sure what the behaviour will be with differently sized first dimensions, but its possible they get screwed up (i.e. data from `x[0]` and `x[1]` are no longer aligned)

Comment (gabrieltseng) [reply src/models/data.py]: Plus, it makes it way easier to pass the arrays to the models (see below)

File: src/models/data.py [hunk 16]
```diff
@@ -180,6 +180,32 @@ def calculate_normalizing_array(self, data_vars: List[str]):
             'std': np.vstack(std).swapaxes(0, 1)
         })
 
+    @staticmethod
+    def move_target_var_to_end_of_dataset(x: xr.Dataset, y: xr.Dataset) -> xr.Dataset:
```
tommylees112: I added this because it allows us to be sure that the final array (index = `-1`) of the numpy array is the target variable when we subselect the `current` month data for all of the variables we are interested in

src/models/regression.py
```diff
@@ -70,7 +70,8 @@ def train(self, num_epochs: int = 1,
                         batch_x[0].shape[0],
                         batch_x[0].shape[1] * batch_x[0].shape[2])
                     if self.include_pred_month:
-                        pred_months = batch_x[1]
+                        # @GABI does this do what you mean it to?
+                        pred_months = batch_x[1].astype(int)
```
tommylees112: I am having troubles here because the `pred_months_onehot` won't merge with the 
in lines [78:80](https://github.com/esowc/ml_drought/pull/49/commits/d67e34acdddb00cdbde10252d81daf33a5dc7ff1#diff-0ae30feeeb42f997c0ae2d1a3d7a7254L77)
```python
                        x_in = np.concatenate(
                            (x_in, pred_months_onehot), axis=-1
                        )
```

because
```
ValueError: all the input arrays must have same number of dimensions
```

tommylees112:
```python
(Pdb) pred_months_onehot
array([], shape=(25, 0, 14), dtype=float64)

(Pdb) pred_months_onehot.shape
(25, 0, 14)

(Pdb) np.eye(14)[pred_months][:, 1:-1]
array([], shape=(25, 0, 14), dtype=float64)

(Pdb) pred_months
array([[12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1]])

(Pdb) pred_months.shape
(25, 2)

(Pdb) x_in.shape
(25, 70)

(Pdb) x_in = np.concatenate( (x_in, pred_months_onehot), axis=-1)
*** ValueError: all the input arrays must have same number of dimensions
```

tommylees112: @gabrieltseng sorry a bit stuck on this one

src/models/data.py
tommylees112: good point!

tommylees112: Should we not try and merge the current and pred months into one array though?
`(latlons, n_features + 1)` 1 being the pred months vector?

tommylees112: Sorry so do you think we should return :

Current behaviour:
```python
return (final_x, final_x_add), final_y
```

What I understand you mean:
```
return (final_x, final_x_add, final_x_curr), final_y
```

The reason I wanted to turn `final_x_add` into an array of shape `(latlons, (n_features - target_var) + 1)` is that the second index (`[1]`) will then be able to be n dimensional if we define other experiments where other vectors are needed

src/models/regression.py
tommylees112: I think the issue is that the `pred_months` array now has 3 dimensions so when you select only the first and last indices (`np.eye(14)[pred_months][:, 1:-1]`) it removes all of those middle indices which relate to ?_each month_?

tommylees112: Should pred_months be `(25, 2)` or should we use `np.unique(pred_months)` ?

tommylees112:
```python
(Pdb) np.eye(14)[np.unique(pred_months)].shape
(2, 14)

(Pdb) np.eye(14)[np.unique(pred_months)]
array([[0., 1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1., 0.]])

(Pdb) np.eye(14)[np.unique(pred_months)][:, 1:-1]
array([[1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.]])

(Pdb) np.eye(14)[np.unique(pred_months)][:, 1:-1].shape
(2, 12)
```

But that also doesn't work because:

```
(Pdb) np.concatenate((x_in, np.eye(14)[np.unique(pred_months)][:, 1:-1]), axis=-1)
*** ValueError: all the input array dimensions except for the concatenation axis must match exactly
(Pdb) x_in.shape
(25, 70)
(Pdb) np.eye(14)[np.unique(pred_months)][:, 1:-1]
array([[1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.]])
(Pdb) np.eye(14)[np.unique(pred_months)][:, 1:-1].shape
(2, 12)
```

src/models/data.py
gabrieltseng: Couldn't other experiments just return more arrays in that tuple?

src/models/data.py
gabrieltseng: Couldn't you just select the indices using x.data_vars?

in line 226:
```python
relevant_indices = [idx for idx, feat in enumerate(x.data_vars) if feat != target_variable]
current = x_np[:, relevant_indices, -1]
```

src/models/regression.py
gabrieltseng: Why does `pred_months` have 3 dimensions?

tommylees112: So when we have only one months worth of data:
```python
(Pdb) x_in.shape
(25, 36)

(Pdb) batch_x[1].astype(int)
array([12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12,
       12, 12, 12, 12, 12, 12, 12, 12])
(Pdb) batch_x[1].astype(int).shape
(25,)
```

the `pred_months` array is 1D

```python
(Pdb) np.eye(14)[pred_months][:, 1:-1]
array([[0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.]])
(Pdb) np.eye(14)[pred_months][:, 1:-1].shape
(25, 12)
```

But when we have two months of data
```python
(Pdb) batch_x[1].astype(int)
array([[12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1]])
(Pdb) pred_months.shape
(25, 2)
```

This code here creates a 3D `pred_months_onehot`

```python
(Pdb) pred_months_onehot = np.eye(14)[pred_months][:, :, 1:-1]
(Pdb) pred_months_onehot
array([[[0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
        [1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.]],

       [[0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
        [1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.]],

(Pdb) pred_months_onehot.shape
(25, 2, 12)
```

tommylees112: Do we want `pred_months_onehot ` to look like:
```python
array([[1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
...
```
If we have 2 months?

tommylees112: I think the issue is that I don't know how to return the `pred_months` for multiple months:
Currently it is returned from the `chunk_array(train_dataloader[0], train_dataloader[1])[0]` 
as:
```python
(Pdb) batch_x[1].astype(int).shape
(25, 2)
(Pdb) batch_x[1].astype(int)
array([[12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1],
       [12,  1]])
```

If we want hte output to look like above (which i think is what we want) how do I add the arrays along the axis of interest?
```
# e.g.
(Pdb) a1=np.array([0,1,0,0])

(Pdb) a2=np.array([0,0,1,0])

(Pdb) a1
array([0, 1, 0, 0])

(Pdb) a2
array([0, 0, 1, 0])

(Pdb) a_ = np.vstack([a1,a2])

(Pdb) a_
array([[0, 1, 0, 0],
       [0, 0, 1, 0]])

(Pdb) a_.shape
(2, 4)
```

How do I add across that dimennsion 0 -> `(1,4)` or `(4,)`
```python
array([0, 1, 1, 0])
```

src/models/data.py
tommylees112: very good point

src/models/data.py
tommylees112: i can't figure out why but this doesn't seem to work in the code htat i'm testing?

tommylees112: i messed up the indexing

tommylees112: i guess that's one reason why i like the xarray approach. there must be a more elegant way of doing what i am doing here

tests/models/test_data.py
```diff
@@ -12,30 +14,51 @@ class TestBaseIter:
...
+        [(True, True, 'one_month_forecast'),
+         (True, False, 'one_month_forecast'),
+         (False, True, 'one_month_forecast'),
+         (False, False, 'one_month_forecast'),
+         (True, True, 'nowcast'),
+         (True, False, 'nowcast'),
+         (False, True, 'nowcast'),
+         (False, False, 'nowcast')]
+    )
+    def test_ds_to_np(self, tmp_path, normalize, to_tensor, experiment):
```
tommylees112: @gabrieltseng this is the test that is failing :(

tests/models/test_data.py
```diff
@@ -53,37 +75,59 @@ def __init__(self):
...
+                x.precip
+                .sel(time=y.time)
+                .stack(dims=['lat', 'lon'])
+                .values.T
+            )
+            got = x_train_data.current
+            assert expected.shape == got.shape, "should have stacked latlon" \
+                " vars as the first dimension in the current array."
+
+            assert all(expected == got), "" \
```
tommylees112: @gabrieltseng this is the major issue. I am sure i am really close but something somewhere is going wrong. Can you just check my test logic and whether i am selecting the right array to compare to?

I am hoping to get `precip` and the final timestep (`time=-1`) and the values stacked (so a `(25,1)` array and that is what should be the `current` vector. but for some reason the value returned from the DataLoader isnot the same?

Any help very much appreciated sorry for the pain!

src/models/data.py
gabrieltseng: What's not working about this?
what error do you get? If we can replace an entire function with a line of code, we should try

gabrieltseng: Saw its fixed; nicely done!

src/models/data.py
```diff
@@ -231,13 +321,15 @@ def __next__(self) -> Tuple[Tuple[Union[np.ndarray, torch.Tensor],
                                 Union[np.ndarray, torch.Tensor]]:
```
gabrieltseng: mypy might throw a fuss here (since we are now returning a tuple with 3 arrays / tensors. The return should be replaced with

` -> Tuple[Tuple[Union[np.ndarray, torch.Tensor], ...], Union[np.ndarray, torch.Tensor]]`

The ellipses tell mypy the Tuple can have any length

tommylees112: Perfect

src/models/linear_network.py
```diff
@@ -114,7 +131,19 @@ def train(self, num_epochs: int = 1,
             for x, y in train_dataloader:
                 for x_batch, y_batch in chunk_array(x, y, batch_size, shuffle=True):
                     optimizer.zero_grad()
-                    pred = self.model(x_batch[0], self._one_hot_months(x_batch[1]))
+                    if self.experiment == 'nowcast':
+                        current = x_batch[2].unsqueeze(1)
```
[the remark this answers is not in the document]
tommylees112: this way the data we feed in will have all of the historical and current data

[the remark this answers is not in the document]
tommylees112: or should I be trying to complete this in the `LinearModel`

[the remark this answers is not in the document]
gabrieltseng: I think this makes more sense as an extra argument being passed to the LinearModel.

The concatenation can happen there with all the other stuff. You would need to pass the experiment to the linear model too, so it knows to concatenate (same as with the pred months stuff)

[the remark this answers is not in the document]
tommylees112: [# L246](https://github.com/esowc/ml_drought/pull/49/files#diff-440d519dccc1eb3b35af18f550d0eb89R246)
```layer_sizes.insert(0, input_size)```

Would I want to just add another `n_variables` to the layer_sizes?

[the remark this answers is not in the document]
Comment (gabrieltseng) [reply src/models/linear_network.py]: Yes, but don't hardcode it. `input_size` is calculated in the train function; add it there.

File: src/models/data.py [hunk 22]
```diff
@@ -148,13 +157,15 @@ def __init__(self, loader: DataLoader) -> None:
...
         self.to_tensor = loader.to_tensor
+        self.experiment = loader.experiment
 
         if self.shuffle:
             # makes sure they are shuffled every epoch
             shuffle(self.data_files)
 
         self.normalizing_dict = loader.normalizing_dict
         self.normalizing_array: Optional[Dict[str, np.ndarray]] = None
+        self.current_normalizing_array: Optional[Dict[str, np.ndarray]] = None
```
gabrieltseng: It looks like the `current_normalizing_array` isn't used. Can it be deleted (from here and from line 189)?

src/models/data.py
```diff
@@ -284,20 +369,39 @@ def __next__(self) -> Tuple[Tuple[Union[np.ndarray, torch.Tensor],
                     cur_max_idx = min(cur_max_idx + 1, self.max_idx)
 
                 out_x.append(arrays.x.historical)
+
+                out_x_curr.append(arrays.x.current)
                 out_x_add.append(arrays.x.pred_months)
+                out_x_add = [x_add for x_add in out_x_add if x_add is not None]
```
gabrieltseng: What is the purpose of this?

gabrieltseng: Can this filtering happen once, at the end? Or something like

```python
if arrays.x.current is not None:
    out_x_curr.append(arrays.x.current)
```

This prevents us from having to loop through the list of `out_x_add` and `out_x_curr` for every file we load.

src/models/neural_networks/linear_network.py
```diff
@@ -81,14 +96,27 @@ def init_weights(self):
...
+        # concatenate the one_hot_month matrix onto X
         if self.include_pred_month:
+            # flatten the array
+            pred_month = pred_month.contiguous().view(x.shape[0], -1)
             x = torch.cat((x, pred_month), dim=-1)
+
+        # concatenate the non-target variables onto X
+        if self.experiment == 'nowcast':
+            assert current is not None
+            current = current.contiguous().view(x.shape[0], -1)
```
gabrieltseng: Shouldn't `current` have shape `[batch_size, dimensions]` anyway? Is this reshaping necessary?

Needs to be changed in the RNN too if not - I just copied the code from here over to there for the tests to pass.

src/models/neural_networks/rnn.py
```diff
@@ -116,6 +128,11 @@ def forward(self, x, pred_month=None):
 
         x = hidden_state.squeeze(0)
 
+        if self.experiment == 'nowcast':
+            assert current is not None
+            current = current.contiguous().view(x.shape[0], -1)
```
gabrieltseng: See comment in the linear model

src/models/utils.py
```diff
@@ -44,7 +43,7 @@ def _chunk_ndarray(x: Tuple[np.ndarray, ...], y: np.ndarray,
                    num_sections: int,
                    shuffle: bool) -> Iterable[Tuple[Tuple[np.ndarray, ...], np.ndarray]]:
 
-    split_x = [np.array_split(x_section, num_sections) for x_section in x]
+    split_x = [np.array_split(x_section, num_sections) for x_section in x if x_section is not None]
```
gabrieltseng: chunk will return a different sized Tuple than it receives, without telling the user that it does that

gabrieltseng: This should be fixed since the indexing will get screwed up depending on what arguments are passed to the model

tests/models/neural_networks/test_linear_network.py
```diff
@@ -44,18 +50,41 @@ def mocktrain(self):
...
-        norm_dict = {'VHI': {'mean': np.zeros(x.to_array().values.shape[:2]),
-                             'std': np.ones(x.to_array().values.shape[:2])}
-                     }
-        with (tmp_path / 'features/normalizing_dict.pkl').open('wb') as f:
+        test_features = tmp_path / f'features/{experiment}/train/hello'
+        test_features.mkdir(parents=True, exist_ok=True)
+
+        # if nowcast we need another x feature
+        if experiment == 'nowcast':
+            x_add1, _, _ = _make_dataset(size=(5, 5), const=True, variable_name='precip')
```
gabrieltseng: There's no reason multi-variable testing should be restricted to the nowcasting case

tests/models/neural_networks/test_linear_network.py
```diff
@@ -65,31 +94,70 @@ def test_train(self, tmp_path, capsys, use_pred_months):
...
 
         model = LinearNetwork(data_folder=tmp_path, layer_sizes=layer_sizes,
-                              dropout=dropout, include_pred_month=use_pred_months)
+                              dropout=dropout, experiment=experiment,
+                              include_pred_month=use_pred_months)
+
         model.train()
 
+        # check the number of input features is properly initialised
+        n_input_features = [
```
gabrieltseng: Nice, this is clever

tests/models/neural_networks/test_linear_network.py
```diff
@@ -65,31 +94,70 @@ def test_train(self, tmp_path, capsys, use_pred_months):
...
-        test_features = tmp_path / 'features/test/hello'
+        test_features = tmp_path / f'features/{experiment}/test/hello'
         test_features.mkdir(parents=True)
 
-        norm_dict = {'VHI': {'mean': np.zeros(x.to_array().values.shape[:2]),
-                             'std': np.ones(x.to_array().values.shape[:2])}
-                     }
-        with (tmp_path / 'features/normalizing_dict.pkl').open('wb') as f:
+        # if nowcast we need another x feature
+        if experiment == 'nowcast':
```
gabrieltseng: Same comment as above; this doesn't need to be in the `if 'nowcast'` statement

tests/models/test_data.py
```diff
@@ -72,41 +95,138 @@ def __init__(self):
...
-                assert target_y == y_np[idx, 0], \
-                    f'Got y different values for lat: {lat}, ' \
-                    f'lon: {lon}.Expected {target_y}, got {y_np[idx, 0]}'
+                    assert target == x_train_data.historical[idx, time, 0], \
+                        'Got different x values for time idx:'\
+                        f'{time}, lat: {lat}, lon: {lon} Expected {target}, '\
+                        f'got {x_train_data.historical[idx, time, 0]}'
+
+    @pytest.mark.parametrize('normalize', [True, False])
+    def test_ds_to_np_multi_vars(self, tmp_path, normalize):
```
gabrieltseng: I think this could be absorbed into the test above. Is anything tested in the single variable test that isn't tested in the multi variable case?

tests/models/test_regression.py
```diff
@@ -25,33 +26,59 @@ def coef_(self):
...
         test_features.mkdir(parents=True)
-
-        norm_dict = {'VHI': {'mean': np.zeros(x.to_array().values.shape[:2]),
-                             'std': np.ones(x.to_array().values.shape[:2])}
-                     }
-        with (tmp_path / 'features/normalizing_dict.pkl').open('wb') as f:
+        pred_features = tmp_path / f'features/{experiment}/test/hello'
+        pred_features.mkdir(parents=True)
+
+        if experiment == 'nowcast':
```
gabrieltseng: Same comment as above

src/models/data.py
tommylees112: yup! good spot

src/models/data.py
tommylees112: thanks!

src/models/neural_networks/linear_network.py
tommylees112: yes most likley - i've just copied that down from before for `pred_months` etc. but actially it's not necessary

src/models/utils.py
tommylees112: Sorry I don't understand this one? What do I need to change here? To tell the user or to have some assert statements to make sure everything is still in the same dimensions?

tests/models/test_regression.py
tommylees112: the nowcast requires other variables though becuase otherwise the `current` array will be empty

tests/models/test_data.py
tommylees112: I think it's important to have a multi-variable test case for 2 reasons
1) we will be using multi variable experiments most of the time
2) the dimensions of the multi-variable arrays will be different and can we not imagine a scenario where the tests pass for the single variable case but fail for the multi-variable case?

tests/models/neural_networks/test_linear_network.py
tommylees112: good point

tests/models/test_data.py
tommylees112: ohhhh you mean to incorporate into the tests above. Good point!

tommylees112: okay done!

src/models/utils.py
tommylees112: or to change the chunking itself?

tests/models/test_data.py
```diff
@@ -72,41 +98,101 @@ def __init__(self):
...
+            assert isinstance(y_np, np.ndarray)
+
+        expected_features = 4 if surrounding_pixels is None else 4 * 9
+        assert x_train_data.historical.shape[-1] == expected_features, "" \
+            "There should be" \
+            "4 historical variables (the final dimension):" \
+            f"{x_train_data.historical.shape}"
+
+        if experiment == 'nowcast':
+            expected_shape = (25, 3) if surrounding_pixels is None else (9, 3 * 9)
```
tommylees112: @gabrieltseng is this returning the expected shape you are expecting?

9 latitude/longitudes and 27 features (3 x 9 latlons)?

Why has the number of `lat lons` changed in the first dimension? Is this the behaviour we expect from `surrounding pixels`?

I would have expected slightly different: `(25, 3*9)` because we have 25 latlons and then 9 surrounding pixels each with 3 features ?

src/models/utils.py
```diff
@@ -33,17 +33,24 @@ def chunk_array(x: Union[Tuple[Union[torch.Tensor, np.ndarray], ...],
...
+                   shuffle: bool) -> Iterable[Tuple[Tuple[Optional[np.ndarray], ...],
+                                                    np.ndarray]]:
 
-    split_x = [np.array_split(x_section, num_sections) for x_section in x if x_section is not None]
+    split_x = []
+    for idx, x_section in enumerate(x):
+        if x_section is not None:
+            split_x.append(np.array_split(x_section, num_sections))
+        else:
+            split_x.append([None] * num_sections)
```
tommylees112: basically we copy the [None] `num_sections` times in order to maintain the structure of the `return_arrays`. I like it

tests/models/test_data.py
gabrieltseng: the `latlon` array only returns the latitude and longitude of the target pixel. 

The "spatial" data of the surrounding pixels is currently encoded in the variable names - this is probably not the best way to do it though

src/models/utils.py
gabrieltseng: Fixed!

PR #53, opened 2019-07-08
Add prediction month to the models

gabrieltseng:
Adds the one hot encoded prediction month to the models, so they know which month they are predicting.

In addition, extends the chunker to handle an arbitrary number of x tensors / arrays.

### DIFF
--- a/scripts/models.py
+++ b/scripts/models.py
@@ -38,10 +38,12 @@ def regression():
                                     shuffle_data=False, mode='test')
     key, val = list(next(iter(test_arrays_loader)).items())[0]
 
-    explanations = predictor.explain(val.x)
-    print(explanations.shape)
-    np.save('shap_regression.npy', explanations)
-    np.save('shap_x.npy', val.x)
+    explain_hist, explain_add = predictor.explain([val.x.historical, val.x.additional])
+    print(explain_hist.shape)
+    np.save('shap_regression_historical.npy', explain_hist)
+    np.save('shap_regression_add.npy', explain_add)
+    np.save('shap_x_hist.npy', val.x.historical)
+    np.save('shap_x_add.npy', val.x.additional)
 
     with open('variables.txt', 'w') as f:
         f.write(str(val.x_vars))
@@ -52,7 +54,7 @@ def regression():
 
     for variable in val.x_vars:
         plt.clf()
-        plot_shap_values(val.x[0], explanations[0], val.x_vars, normalizing_dict, variable,
+        plot_shap_values(val.x[0], explain_hist[0], val.x_vars, normalizing_dict, variable,
                          normalize_shap_plots=True, show=False)
         plt.savefig(f'{variable}_linear_regression.png', dpi=300, bbox_inches='tight')
 
@@ -73,10 +75,13 @@ def linear_nn():
                                     shuffle_data=False, mode='test', to_tensor=True)
     key, val = list(next(iter(test_arrays_loader)).items())[0]
 
-    explanations = predictor.explain(val.x[:3])
-    print(explanations.shape)
-    np.save('shap_linear_network.npy', explanations)
-    np.save('shap_x_linear_network.npy', val.x[:3])
+    explain_hist, explain_add = predictor.explain([val.x.historical[:3], val.x.additional[:3]])
+    print(explain_hist.shape)
+    np.save('shap_linear_network_hist.npy', explain_hist)
+    np.save('shap_linear_network_add.npy', explain_add)
+
+    np.save('shap_x_linear_network_hist.npy', val.x.historical[:3])
+    np.save('shap_x_linear_network_add.npy', val.x.additional[:3])
 
     with open('variables_linear_network.txt', 'w') as f:
         f.write(str(val.x_vars))
@@ -87,7 +92,7 @@ def linear_nn():
 
     for variable in val.x_vars:
         plt.clf()
-        plot_shap_values(val.x[0].numpy(), explanations[0], val.x_vars, normalizing_dict, variable,
+        plot_shap_values(val.x[0].numpy(), explain_hist[0], val.x_vars, normalizing_dict, variable,
                          normalize_shap_plots=True, show=False)
         plt.savefig(f'{variable}_linear_network.png', dpi=300, bbox_inches='tight')
 
--- a/scripts/models.py
+++ b/scripts/models.py
@@ -54,8 +54,8 @@ def regression():
 
     for variable in val.x_vars:
         plt.clf()
-        plot_shap_values(val.x[0], explain_hist[0], val.x_vars, normalizing_dict, variable,
-                         normalize_shap_plots=True, show=False)
+        plot_shap_values(val.x.historical[0], explain_hist[0], val.x_vars, normalizing_dict,
+                         variable, normalize_shap_plots=True, show=False)
         plt.savefig(f'{variable}_linear_regression.png', dpi=300, bbox_inches='tight')
 
 
--- a/scripts/models.py
+++ b/scripts/models.py
@@ -38,12 +38,12 @@ def regression():
                                     shuffle_data=False, mode='test')
     key, val = list(next(iter(test_arrays_loader)).items())[0]
 
-    explain_hist, explain_add = predictor.explain([val.x.historical, val.x.additional])
+    explain_hist, explain_add = predictor.explain([val.x.historical, val.x.pred_months])
     print(explain_hist.shape)
     np.save('shap_regression_historical.npy', explain_hist)
     np.save('shap_regression_add.npy', explain_add)
     np.save('shap_x_hist.npy', val.x.historical)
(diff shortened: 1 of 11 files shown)


src/models/data.py
```diff
@@ -9,9 +10,15 @@
 from typing import cast, Dict, Optional, Union, List, Tuple
 
 
+@dataclass
+class TrainData:
+    historical: Union[np.ndarray, torch.Tensor]
+    pred_months: Union[np.ndarray, torch.Tensor]
+
```
tommylees112: is the idea here to also have another attribute `additional` or `current_month` perhaps:

I have a line [here](https://github.com/esowc/ml_drought/pull/49/files?file-filters%5B%5D=.py#diff-b07d4def3bfdc11dc410d8848b939e1aR15)
```
current = Union[np.ndarray, torch.Tensor, None]
```

src/models/data.py
```diff
@@ -169,6 +177,10 @@ def ds_folder_to_np(self,
         # first, x
         x_np = x_np.reshape(x_np.shape[0], x_np.shape[1], x_np.shape[2] * x_np.shape[3])
         x_np = np.moveaxis(np.moveaxis(x_np, 0, 1), -1, 0)
+        # then, the x month
+        assert len(y.time) == 1, f'Expected y to only have 1 timestamp! Got {len(y.time)}'
+        target_month = datetime.strptime(str(y.time.values[0])[:-3], '%Y-%m-%dT%H:%M:%S.%f').month
+        x_months = np.array([target_month] * x_np.shape[0])
```
tommylees112: doesn't this just create a vector of the `target_month` rather than the months of the input timesteps ? or is this what we want? 

I.e. to know that we are looking at predicting June vs. to know that we are using March, April, May to predict June? Or is this something implicitly picked up?

src/models/data.py
```diff
@@ -200,27 +214,31 @@ def ds_folder_to_np(self,
...
 class _TrainIter(_BaseIter):
 
-    def __next__(self) -> Tuple[Union[np.ndarray, torch.Tensor],
+    def __next__(self) -> Tuple[Tuple[Union[np.ndarray, torch.Tensor],
+                                      Union[np.ndarray, torch.Tensor]],
                                 Union[np.ndarray, torch.Tensor]]:
 
         if self.idx < self.max_idx:
-            out_x, out_y = [], []
+            out_x, out_x_add, out_y = [], [], []
```
tommylees112: this is where we have the power with the `additional` vectors? (i.e. for the `current` timesteps in the `nowcast`)

src/models/data.py
```diff
@@ -229,17 +247,20 @@ def __next__(self) -> Tuple[Union[np.ndarray, torch.Tensor],
...
             final_x = np.concatenate(out_x, axis=0)
+            final_x_add = np.concatenate(out_x_add, axis=0)
             final_y = np.concatenate(out_y, axis=0)
             if final_x.shape[0] == 0:
                 raise StopIteration()
             if self.to_tensor:
-                return torch.from_numpy(final_x).float(), torch.from_numpy(final_y).float()
-            return final_x, final_y
+                return (torch.from_numpy(final_x).float(),
+                        torch.from_numpy(final_x_add).float()), torch.from_numpy(final_y).float()
```
tommylees112: what if there is no additional data (as in our vanilla models?)

src/models/utils.py
```diff
@@ -1,21 +1,23 @@
 import torch
 import numpy as np
 from random import shuffle as shuffle_list
-from typing import Iterable, Union, Tuple
+from typing import cast, Iterable, Union, Tuple
 
 
-def chunk_array(x: Union[torch.Tensor, np.ndarray],
+def chunk_array(x: Union[Tuple[Union[torch.Tensor, np.ndarray], ...],
```
tommylees112: this is super helpful thankyou Gabi!

src/models/utils.py
```diff
@@ -27,31 +29,37 @@ def chunk_array(x: Union[torch.Tensor, np.ndarray],
...
-    split_x, split_y = np.array_split(x, num_sections), np.array_split(y, num_sections)
-    return_arrays = list(zip(split_x, split_y))
+    split_x = [np.array_split(x_section, num_sections) for x_section in x]
+    split_y = np.array_split(y, num_sections)
+    return_arrays = list(zip(*split_x, split_y))
 
     if shuffle:
         shuffle_list(return_arrays)
-    return return_arrays
+    return [(chunk[:-1], chunk[-1]) for chunk in return_arrays]  # type: ignore
```
tommylees112: this returns separate chunks for the additional / final vector of `current` data?

src/models/parsimonious.py
```diff
@@ -36,7 +36,7 @@ def predict(self) -> Tuple[Dict[str, Dict[str, np.ndarray]], Dict[str, np.ndarra
                     print('Target variable not in prediction data!')
                     raise e
 
-                preds_dict[key] = val.x[:, -1, [target_idx]]
+                preds_dict[key] = val.x.historical[:, -1, [target_idx]]
```
tommylees112: so this model ONLY has `historical` information right?

src/models/linear_network.py
```diff
@@ -167,20 +176,35 @@ def _get_background(self, sample_size: int = 100) -> torch.Tensor:
...
+                    if self.include_pred_month:
+                        output_pred_months.append(self._one_hot_months(x[1][idx: idx + 1]))
+        if self.include_pred_month:
+            return [torch.stack(output_tensors), torch.cat(output_pred_months, dim=0)]
+        else:
+            return torch.stack(output_tensors)
+
+    @staticmethod
+    def _one_hot_months(indices: torch.Tensor) -> torch.Tensor:
+        return torch.eye(14)[indices.long()][:, 1:-1]
```
tommylees112: what is the identity matrix doing here? is the output a `14 x 12 matrix`?

scripts/models.py
```diff
@@ -38,10 +38,12 @@ def regression():
                                     shuffle_data=False, mode='test')
     key, val = list(next(iter(test_arrays_loader)).items())[0]
 
-    explanations = predictor.explain(val.x)
-    print(explanations.shape)
-    np.save('shap_regression.npy', explanations)
-    np.save('shap_x.npy', val.x)
+    explain_hist, explain_add = predictor.explain([val.x.historical, val.x.pred_months])
```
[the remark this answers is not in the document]
gabrieltseng: Ah; the variable name is misleading. Its explain historical, not histogram.

src/models/data.py
gabrieltseng: yep! This can be extended as necessary

src/models/data.py
gabrieltseng: Since everything is flattened, it wouldn't really make sense to input the prediction months (you'd end up with a list of integers); I thought this would be better captured just by adding the prediction month.

(So yes; I am expecting the input months to be implicitly picked up). We can change this if it doesn't seem to work

src/models/data.py
gabrieltseng: yes

src/models/utils.py
gabrieltseng: It returns a tuple for all the x data vectors, and the y data.

This allows you to iterate using:

```python
for x, y in dataloader:
    ...
```
where x is now a tuple of arrays / tensors

src/models/parsimonious.py
gabrieltseng: right (since its predictions are the VHI for the previous month).

src/models/linear_network.py
gabrieltseng: The output is a `indices.shape[0] * 12` matrix.

You take the identity matrix, and take the indices in the `indices` array; this is the one hot encoding.

We ignore the 0th (since there is no 0 month) and the extra class.

src/models/data.py
gabrieltseng: The iterator will always return the pred months, but the models can decide not to use them (if `include_pred_month` is False)

gabrieltseng:
Re: the experiment name, I think maybe we could think about that after your stuff is added?

But I agree that this is a pretty universal use case, so it might not be worth adding an extra folder just for this.

One additional note:

The deep explainer is currently failing with 

```
File "/home/gabrieltseng95/miniconda3/envs/esowc-drought/lib/python3.7/site-packages/shap/explainers/deep/deep_pytorch.py", line 107, in <listcomp>
    grads = [torch.autograd.grad(selected, x)[0].cpu().numpy() for x in X]
  File "/home/gabrieltseng95/miniconda3/envs/esowc-drought/lib/python3.7/site-packages/torch/autograd/__init__.py", line 149, in grad
    inputs, allow_unused)
RuntimeError: Trying to backward through the graph a second time, but the buffers have already been freed. Specify retain_graph=True when calling backward the first time.
```

I am not sure if this is a bug in this code, or in shap. I'll take a look, but merge this first to not block your stuff.

gabrieltseng:
Turns out its a problem with shap; I'll try and fix it ASAP

PR #57, opened 2019-07-11
Monthly mean preprocessor for ERA5 data

gabrieltseng:
A few tweaks to the exporter too


src/exporters/cds.py
```diff
@@ -249,7 +250,7 @@ def export(self,
                granularity: str = 'hourly',
                show_api_request: bool = True,
                selection_request: Optional[Dict] = None,
-               break_up: bool = True,
+               break_up: bool = False,
```
tommylees112: good point!

PR #58, opened 2019-07-12
Add an RNN

gabrieltseng:
Adds a Recurrent Neural Network

### DIFF
--- a/scripts/models.py
+++ b/scripts/models.py
@@ -7,7 +7,7 @@ import matplotlib.pyplot as plt
 import pickle
 
 from src.analysis import plot_shap_values
-from src.models import Persistence, LinearRegression, LinearNetwork
+from src.models import Persistence, LinearRegression, LinearNetwork, RecurrentNetwork
 from src.models.data import DataLoader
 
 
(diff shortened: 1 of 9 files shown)


src/models/neural_networks/rnn.py
```diff
@@ -0,0 +1,175 @@
...
+    """
+
+    def __init__(self, input_size, hidden_size, batch_first=True):
+        super().__init__()
+
+        self.input_size = input_size
+        self.hidden_size = hidden_size
+        self.batch_first = batch_first
+
+        self.forget_gate = nn.Sequential(*[
```
tommylees112: what's the initial `*` doing?

```
self.forget_gate = nn.Sequential(*[
```

src/models/neural_networks/rnn.py
```diff
@@ -0,0 +1,175 @@
...
+
+        self.output_gate = nn.Sequential(*[
+            nn.Linear(in_features=input_size + hidden_size, out_features=hidden_size,
+                      bias=True), nn.Sigmoid()
+        ])
+
+        self.cell_state_activation = nn.Tanh()
+
+    def forward(self, x, state):
+        hidden, cell = state
```
tommylees112: what is the type of `state`? Is `state` the current information flowing through the forget/update gates?

tommylees112: Screenshot 2019-07-16 at 10 02 31

Is it the `C` here?
From http://colah.github.io/posts/2015-08-Understanding-LSTMs/

src/models/neural_networks/rnn.py
```diff
@@ -0,0 +1,175 @@
...
+
+        if self.include_pred_month:
+            x = torch.cat((x, pred_month), dim=-1)
+        for layer_number, dense_layer in enumerate(self.dense_layers):
+            x = dense_layer(x)
+        return x
+
+
+class UnrolledRNN(nn.Module):
+    """An unrolled RNN. The motivation for this is mainly so that we can explain this model using
```
tommylees112: what does unrolled mean?

src/models/neural_networks/rnn.py
```diff
@@ -0,0 +1,175 @@
...
+
+        for dense_layer in self.dense_layers:
+            nn.init.kaiming_uniform_(dense_layer.weight.data)
+            nn.init.constant_(dense_layer.bias.data, 0)
+
+    def forward(self, x, pred_month=None):
+
+        sequence_length = x.shape[1]
+
+        hidden_state = torch.zeros(1, x.shape[0], self.hidden_size)
```
tommylees112: What's the difference between the `hidden state` and `cell_State`?

Screenshot 2019-07-16 at 10 04 46


Is it the difference between `A` (the layers) and `C` the conveyor belt passing through the layers?

src/models/neural_networks/rnn.py
tommylees112: See above: `The reason the RNN is unrolled here is to apply dropout to each timestep `

scripts/models.py
```diff
@@ -67,37 +67,57 @@ def linear_nn():
...
+    # if the working directory is alread ml_drought don't need ../data
+    if Path('.').absolute().as_posix().split('/')[-1] == 'ml_drought':
+        data_path = Path('data')
+    else:
+        data_path = Path('../data')
 
-    # plot the variables
-    with (data_path / 'features/normalizing_dict.pkl').open('rb') as f:
-        normalizing_dict = pickle.load(f)
+    predictor = RecurrentNetwork(hidden_size=128,
```
tommylees112: So this will create a RNN with 
[ input_size -> 100 ] * 128 layers?

src/models/neural_networks/rnn.py
gabrieltseng: [`nn.Sequential`](https://pytorch.org/docs/stable/nn.html?highlight=sequential#torch.nn.Sequential) takes as input all of the Modules which will make up the container (not a list of the Modules) - the `*` takes the list and passes the elements of the list as arguments.

src/models/neural_networks/rnn.py
gabrieltseng: State is a tuple `(hidden, cell)` (see line 157) - so its `(h_{t-1}, C_{t-1})`

src/models/neural_networks/rnn.py
gabrieltseng: Note the two arrows coming into the cell from the previous cell - one represents the cell state, the other the hidden state.

The cell state is intended to represent the "long term" information being captured. The hidden state is the output of the specific cell - `h_t`. Note that this is both the output of the cell, and is also passed to the next cell.

scripts/models.py
gabrieltseng: There is only one RNN layer, but its hidden state has 128 dimensions (so the output of each RNN cell will have shape `[batch_size, 128]`.

Then, the model takes the last hidden state, and puts it through a linear model with layer sizes `[128, 1]`.

PR #59, opened 2019-07-15
Include surrounding pixels in the model

gabrieltseng:
This PR allows a 'surrounding pixel' radius to be defined. If this happens, then the pixels around the target pixel are included in the training data.


src/models/data.py
```diff
@@ -160,9 +167,13 @@ def calculate_normalizing_array(self, data_vars: List[str]):
         self.normalizing_dict = cast(Dict[str, Dict[str, np.ndarray]], self.normalizing_dict)
 
         mean, std = [], []
+        normalizing_dict_keys = self.normalizing_dict.keys()
         for var in data_vars:
-            mean.append(self.normalizing_dict[var]['mean'])
-            std.append(self.normalizing_dict[var]['std'])
+            for norm_var in normalizing_dict_keys:
+                if var.endswith(norm_var):
```
tommylees112: what is this doing?

src/models/data.py
```diff
@@ -234,6 +247,18 @@ def ds_folder_to_np(self,
         return ModelArrays(x=x_data, y=y_np, x_vars=list(x.data_vars),
                            y_var=list(y.data_vars)[0])
 
+    @staticmethod
+    def _add_surrounding(x: xr.Dataset, surrounding_pixels: int) -> xr.Dataset:
```
tommylees112: So in the `dataloader` this PR adds a new variable which is the `shifted_varname` so we have `lat1_lon1_VHI` for the historical months and `lat1_lon1_precip` for the historical months.

Does you expect it will work with the `nowcast` experiment?

tests/models/test_linear_network.py
```diff
@@ -25,7 +26,8 @@ def mocktrain(self):
         monkeypatch.setattr(LinearNetwork, 'train', mocktrain)
 
         model = LinearNetwork(data_folder=tmp_path, layer_sizes=layer_sizes,
-                              dropout=dropout, include_pred_month=include_pred_month)
+                              dropout=dropout, include_pred_month=include_pred_month,
+                              surrounding_pixels=surrounding_pixels)
```
tommylees112: How big will the number of features be when we have a `surrounding_pixels` argument of 1?

with 3 variables:
```
( (12 months x 3) x 2 latlon shifts (surrounding_pixels =1) )  + 12 one_hot_months 
```

src/models/data.py
tommylees112:
```
                if var.endswith(norm_var):
```

gabrieltseng: Now, we have variables that no longer explicitly match what is in the normalizing dict. For instance, instead of `precip` we have `lat_1_lon_1_precip`. This 'shifted precip' will have the same mean and std, so should be normalized using the same values in the normalizing dict - this achieves that.

src/models/data.py
gabrieltseng: It should - `add_surrounding` is called in `ds_to_np` before the data is split into historical and current, so the current should also have the surrounding pixels.

The only thing that will need to change is that we will need to use the `endswith` check to eliminate VHI, since an exact match will miss the `lat_1_lon_1_VHI` variables.

tests/models/test_linear_network.py
gabrieltseng: So if `surrounding pixels == 1`, historical gets `9x` more features. So I think it would be:

`12 months x 3 variables x 9 pixels) + 12 one hot months`

PR #60, opened 2019-07-15
Make the training data 12 months by default

gabrieltseng:


gabrieltseng:
The plotting function assumes 11 months - needs to be updated to figure out how many months from the input data

tommylees112:
> The plotting function assumes 11 months - needs to be updated to figure out how many months from the input data

Is this to be done or done already?

PR #64, opened 2019-07-18
[minor] Add nonlinearities to the final dense layers of the RNN

gabrieltseng:


src/models/neural_networks/rnn.py
```diff
@@ -84,15 +89,16 @@ def __init__(self, features_per_month, dense_features, hidden_size,
             assert current_size is not None
             dense_input_size += current_size
         dense_features.insert(0, dense_input_size)
-        if dense_features[-1] != 1:
-            dense_features.append(1)
 
         self.dense_layers = nn.ModuleList([
-            nn.Linear(in_features=dense_features[i - 1],
-                      out_features=dense_features[i])
+            LinearBlock(in_features=dense_features[i - 1],
```
tommylees112: what is the difference between:
`nn.Linear` and `LinearBlock` ?

gabrieltseng: The Linear Block is defined here: https://github.com/esowc/ml_drought/blob/models/rnn/src/models/neural_networks/base.py#L218

It introduces the nonlinearities (e.g. the ReLU activation) that allows the linear layers to express non-linear functions.

gabrieltseng:
The final layer isn't nonlinear, but the linear blocks allow the linear layers to express non linear functions (instead of just having vanilla linear layers)

tommylees112:
Gotcha!

PR #66, opened 2019-07-19
Fix validation bug in the nowcaster

gabrieltseng:
Opening a PR to leave comments!


src/models/neural_networks/base.py
```diff
@@ -129,7 +129,15 @@ def train(self, num_epochs: int = 1,
                 val_rmse = []
                 with torch.no_grad():
                     for x, y in val_dataloader:
-                        val_pred_y = self.model(x[0], self._one_hot_months(x[1]))
+                        if self.experiment == 'nowcast':
+                            current = x[2]
+                            val_pred_y = self.model(
+                                x[0], self._one_hot_months(x[1]), current
```
gabrieltseng: @tommylees112 , is this `if` statement necessary?

Sorry I didn't catch it before, but shouldn't the model not care if current is passed - it knows to ignore it if the experiment isn't `nowcast`.

It's similar to how the one hot months are passed to the model whether or not the model will use them, and the model decides whether or not to concatenate them to `x`.

We could reduce 10 lines of code from each model training function.

gabrieltseng:
As another note, the `current` array isn't being included in `Regression.calculate_big_mean`, which will cause problems when we try to use shap values in the nowcasting experiment - it should be added there too.

src/models/neural_networks/base.py
gabrieltseng: Fixed in https://github.com/esowc/ml_drought/pull/66/commits/19d7b41e1b95d2dbe5c40e96be85bb6126d12e63

src/models/regression.py
```diff
@@ -235,6 +235,9 @@ def _calculate_big_mean(self) -> np.ndarray:
                 # for us its + 2, since 0 is not a class either
                 pred_months_onehot = np.eye(14)[pred_months][:, 1:-1]
                 x_in = np.concatenate((x_in, pred_months_onehot), axis=-1)
+            if self.experiment == 'nowcast':
```
tommylees112: What does the `_calculate_big_mean()` function actually do? Calculate the mean across the whole dataset (incl. the current timestep??) to act as a baseline for the calculation of SHAP values?

gabrieltseng: It calculates the variable-wise mean of the training data. Yes, this is to act as a baseline for the shap values (in the `explain` function, you can see that the output of this function is passed to the linear explainer).

PR #67, opened 2019-07-19
Make it easy to save and load models

gabrieltseng:
Makes it easy to save and load models. Mainly, this is achieved in two ways:

1) A uniform format is used when saving the models
2) A [model loading](https://github.com/esowc/ml_drought/blob/models/saving_and_loading/src/models/__init__.py#L17) function is implemented, which leverages the uniform format to return a pretrained model

~~Tests to make sure all of this works as desired need to be implemented~~

### DIFF
--- a/tests/models/test_regression.py
+++ b/tests/models/test_regression.py
@@ -41,8 +41,9 @@ class TestLinearRegression:
 
         with (tmp_path / 'models/one_month_forecast/linear_regression/model.pkl').open('rb') as f:
             model_dict = pickle.load(f)
-        assert np.array_equal(coef_array, model_dict['coef']), 'Different coef array saved!'
-        assert np.array_equal(intercept_array, model_dict['intercept']), \
+        assert np.array_equal(coef_array, model_dict['model']['coef']), \
+            'Different coef array saved!'
+        assert np.array_equal(intercept_array, model_dict['model']['intercept']), \
             'Different intercept array saved!'
         assert model_dict['experiment'] == 'one_month_forecast', 'Different experiment saved!'
 
--- a/tests/models/test_regression.py
+++ b/tests/models/test_regression.py
@@ -14,13 +14,18 @@ class TestLinearRegression:
 
     def test_save(self, tmp_path, monkeypatch):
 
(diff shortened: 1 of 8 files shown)


tests/models/test_regression.py
```diff
@@ -31,11 +36,16 @@ def coef_(self):
         model.save_model()
 
         assert (
-            tmp_path / 'models/one_month_forecast/linear_regression/model.npy'
+            tmp_path / 'models/one_month_forecast/linear_regression/model.pkl'
         ).exists(), f'Model not saved!'
 
-        saved_model = np.load(tmp_path / 'models/one_month_forecast/linear_regression/model.npy')
-        assert np.array_equal(model_array, saved_model), f'Different array saved!'
+        with (tmp_path / 'models/one_month_forecast/linear_regression/model.pkl').open('rb') as f:
```
tommylees112: Why do these tests differ from the tests in `tests/models/__init__.py`

tommylees112: Doesn't this section [(L117:L124)](https://github.com/esowc/ml_drought/blob/d7b6c25c49943149c46d0704de9ac24e7f2a55bd/tests/models/__init__.py#L120) test the same thing?

```python
        new_model = load_model(model_path)
        assert type(new_model) == LinearRegression

        assert model.model.coef_ == coef_array
        assert model.model.intercept_ == intercept_array
        assert new_model.include_pred_month == model.include_pred_month
        assert new_model.experiment == model.experiment
        assert new_model.surrounding_pixels == model.surrounding_pixels
```

gabrieltseng:
Nicely caught - that's a mistake on my part. Thanks! I'll update it

tests/models/test_regression.py
gabrieltseng: So the test in that line is the same - just that the `pkl` file is saved. The reason they are duplicated is because in both steps, its a necessary step for the next tests.

The tests after are slightly different. In `test_regression`, it tests that those attributes are saved in the `pkl` object. In the `__init__.py` file, it tests that those attributes are correct in the loaded model.

PR #72, opened 2019-07-23
Exporter for ESA CCI Landcover data

tommylees112:
Download the ESA CCI Landcover Data from http://maps.elie.ucl.ac.be/CCI/viewer/download.php.

Main functionality is using bash commands:
`wget` and `unzip`

These are wrapped into python functions
`wget_file()` and `unzip()`


tests/exporters/test_esa_cci.py
```diff
@@ -0,0 +1,77 @@
...
+        exporter = ESACCIExporter(tmp_path)
+
+        url_path = 'ftp://geo10.elie.ucl.ac.be/v207/ESACCI-LC-L4-LCCS-Map-'\
+            '300m-P1Y-1992_2015-v2.0.7b.nc.zip'
+        folder = (tmp_path / 'raw' / 'esa_cci_landcover').as_posix()
+        mock_system.assert_called_once_with(
+            f'wget {url_path} -P {folder}'
+        )
+
+    def test_read_legend(self, mock_system, tmp_path, capsys):
```
tommylees112: don't know if this function would even be required? the idea was to write the response of the url request as a string and then get the pandas function to read this string. Needs some more love

PR #74, opened 2019-07-23
Preprocessor for ESA Landcover data

tommylees112:
Preprocessing the landcover data (there is an issue with the regridding on the ouce_server)

        Process:
        -------
        * chop region of interset (ROI)
        * regrid to same spatial grid as a reference dataset (`regrid`)
        * create new dataset with these dimensions
        * assign time stamp
        * Save the output file to new folder


tommylees112:
TODO:

- Need to copy the array forwards JUST FOR THAT YEAR 
- use the `src.utils.create_shape_aligned_climatology` or something similar to copy the array for the start of the year forwards in time
- default behaviour for upsampling is `nearest` which is fine but it means the years are split. Do we prefer that behaviour? 

```python
In [84]: d1 = d.sel(time = '1995-01-31')
In [82]: d2 = d.sel(time = '1995-12-31')

In [85]: (d1 == d2).mean()
Out[85]:
<xarray.Dataset>
Dimensions:   ()
Data variables:
    lc_class  float64 0.9986
```

src/preprocess/base.py
```diff
@@ -48,12 +48,24 @@ def __init__(self, data_folder: Path = Path('data')) -> None:
...
             target_folder = self.raw_folder / self.dataset
         else:
             target_folder = self.interim
-        outfiles = list(target_folder.glob('**/*.nc'))
+
+        if years is not None:
+            assert int(list(target_folder.glob('**'))[0])  # type: ignore
+            outfiles = [
+                f for f in target_folder.glob('**/*.nc')
+                if int(f.parents[0]) in years  # type: ignore
```
gabrieltseng: This assumes the files are exported to a `year/name.nc` format - this is not necessarily the case (for instance, the ESA CCI exporter doesn't do that). 

What was the motivation for this change?

src/preprocess/base.py
```diff
@@ -161,14 +173,17 @@ def chop_roi(ds: xr.Dataset,
...
+                    variable: Optional[str] = None) -> None:
 
         ds = xr.open_mfdataset(self.get_filepaths('interim'))
 
         if resample_time is not None:
             ds = self.resample_time(ds, resample_time, upsampling)
 
-        out = self.out_dir / f'{self.dataset}\
-        {"_" + subset_str if subset_str is not None else ""}.nc'.replace(' ', '')
+        out = self.out_dir / f'{self.dataset}' \
```
gabrieltseng: Same question as above - why is this necessary? Where is it used?

src/utils.py
```diff
@@ -129,6 +130,35 @@ def create_shape_aligned_climatology(ds, clim, variable, time_period):
     return new_clim
 
 
+def get_modal_value_across_time(da: xr.DataArray) -> xr.DataArray:
```
gabrieltseng: Can a test for this be added?

src/preprocess/base.py
tommylees112: to only preprocess a subset of the years (if that option is specifieD).

speeds the preprocessing up when it's being tested on servers etc.

plus means that in future if only want to do a subset of the years you can preprocess them. but should probably be better documented or done elsewhere

src/preprocess/base.py
tommylees112: ah this got muddled with the ERA5land stuff sorry :( 

It was supposed to deal with the fact that there are so many different era5 land variables that itmakse sense to preprocess them separately

src/utils.py
tommylees112: good point

src/preprocess/base.py
gabrieltseng: should we move this over to the era5 land PR?

gabrieltseng: I think what makes more sense is to allow the merge files to be overridden completely, rather than having these new options that don't apply to all preprocessors

src/utils.py
gabrieltseng: Thanks!

PR #75, opened 2019-07-23
Updates to the engineer

gabrieltseng:
Updates to the engineer as I was working on the documentation.

Mainly:

1) Since both engineers have the same API, they are collapsed into a single engineer so they can better be differentiated programmatically
2) Hide a bunch of functions


gabrieltseng:
Yea, no rush - this isn't time sensitive

PR #76, opened 2019-07-23
Add EALSTM 

gabrieltseng:
Adds an entity-aware LSTM, as described here: https://github.com/kratzert/ealstm_regional_modeling

Most of the code is just copied from the EALSTM described there, and fitted to our current RNN model.

Currently, the static data is the latlon data, but this should be updated to include landcover, and eventually topographical data.

### DIFF
--- a/src/models/data.py
+++ b/src/models/data.py
@@ -19,7 +19,7 @@ class TrainData:
     # latlons are repeated here so they can be tensor-ized and
     # normalized
     latlons: Union[np.ndarray, torch.Tensor]
-    yearly_means: Union[np.ndarray, torch.Tensor]
+    yearly_aggs: Union[np.ndarray, torch.Tensor]
 
 
 @dataclass
@@ -185,7 +185,7 @@ class _BaseIter:
         return self
 
     def calculate_normalizing_array(self, data_vars: List[str],
-                                    yearly_mean: bool = False) -> Dict[str, np.ndarray]:
+                                    yearly_agg: bool = False) -> Dict[str, np.ndarray]:
         # If we've made it here, normalizing_dict is definitely not None
         self.normalizing_dict = cast(Dict[str, Dict[str, np.ndarray]], self.normalizing_dict)
 
@@ -196,7 +196,7 @@ class _BaseIter:
                 if var.endswith(norm_var):
                     var_mean = self.normalizing_dict[norm_var]['mean']
                     var_std = self.normalizing_dict[norm_var]['std']
-                    if yearly_mean:
+                    if yearly_agg:
                         var_mean = np.mean(var_mean)
                         var_std = np.mean(var_std)
                     mean.append(var_mean)
@@ -205,9 +205,9 @@ class _BaseIter:
 
         mean_np, std_np = np.vstack(mean), np.vstack(std)
 
-        if yearly_mean:
+        if yearly_agg:
             mean_np, std_np = mean_np.squeeze(1), std_np.squeeze(1)
-        if not yearly_mean:
+        else:
             # swapaxes so that its [timesteps, features], not [features, timesteps]
             mean_np, std_np = mean_np.swapaxes(0, 1), std_np.swapaxes(0, 1)
 
@@ -230,15 +230,20 @@ class _BaseIter:
         yearly_mean = x.mean(dim=['time', 'lat', 'lon'])
         yearly_mean_np = yearly_mean.to_array().values
 
+        yearly_std = x.std(dim=['time', 'lat', 'lon'])
+        yearly_std_np = yearly_std.to_array().values
+
+        yearly_agg = np.concatenate((yearly_mean_np, yearly_std_np))
+
         x = self._add_extra_dims(x, self.surrounding_pixels, self.monthly_aggs)
 
         x_np, y_np = x.to_array().values, y.to_array().values
 
         if (self.normalizing_dict is not None) and (self.normalizing_array is None):
             self.normalizing_array = self.calculate_normalizing_array(
-                list(x.data_vars), yearly_mean=False)
+                list(x.data_vars), yearly_agg=False)
             self.normalizing_array_ym = self.calculate_normalizing_array(
-                list(yearly_mean.data_vars), yearly_mean=True)
+                list(yearly_mean.data_vars) + list(yearly_std.data_vars), yearly_agg=True)
 
         # first, x
         x_np = x_np.reshape(x_np.shape[0], x_np.shape[1], x_np.shape[2] * x_np.shape[3])
@@ -251,7 +256,7 @@ class _BaseIter:
             str(y.time.values[0])[:-3], '%Y-%m-%dT%H:%M:%S.%f'
         ).month
         x_months = np.array([target_month] * x_np.shape[0])
-        yearly_mean_np = np.vstack([yearly_mean_np] * x_np.shape[0])
+        yearly_agg = np.vstack([yearly_agg] * x_np.shape[0])
 
         # then, latlons
         lons, lats = np.meshgrid(x.lon.values, x.lat.values)
@@ -266,7 +271,7 @@ class _BaseIter:
         if (self.normalizing_array is not None) and (self.normalizing_array_ym is not None):
             x_np = (x_np - self.normalizing_array['mean']) / (self.normalizing_array['std'])
             train_latlons = train_latlons / [90, 180]
-            yearly_mean_np = (yearly_mean_np - self.normalizing_array_ym['mean']) / \
+            yearly_agg = (yearly_agg - self.normalizing_array_ym['mean']) / \
                 self.normalizing_array_ym['std']
 
         if self.experiment == 'nowcast':
@@ -281,7 +286,7 @@ class _BaseIter:
                 historical=historical,
                 pred_months=x_months,
                 latlons=train_latlons,
-                yearly_means=yearly_mean_np
+                yearly_aggs=yearly_agg
             )
 
         else:
@@ -290,7 +295,7 @@ class _BaseIter:
                 historical=x_np,
                 pred_months=x_months,
                 latlons=train_latlons,
-                yearly_means=yearly_mean_np
+                yearly_aggs=yearly_agg
             )
 
         assert y_np.shape[0] == x_np.shape[0], f'x and y data have a different ' \
@@ -322,7 +327,7 @@ class _BaseIter:
             train_data.historical = train_data.historical[notnan_indices]
             train_data.pred_months = train_data.pred_months[notnan_indices]  # type: ignore
             train_data.latlons = train_data.latlons[notnan_indices]
-            train_data.yearly_means = train_data.yearly_means[notnan_indices]
+            train_data.yearly_aggs = train_data.yearly_aggs[notnan_indices]
 
             y_np = y_np[notnan_indices]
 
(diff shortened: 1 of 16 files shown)


gabrieltseng:
TODO: add mean values

src/models/neural_networks/ealstm.py
```diff
@@ -0,0 +1,281 @@
...
+            nn.init.kaiming_uniform_(dense_layer.linear.weight.data)
+
+        nn.init.kaiming_uniform_(self.final_dense.weight.data)
+        nn.init.constant_(self.final_dense.bias.data, 0)
+
+    def forward(self, x, pred_month=None, latlons=None, current=None):
+
+        # has to be optional to respect the positionality of the neural
+        # nets
+        assert latlons is not None, "Latlons can't be None!"
```
tommylees112: can we not make latlons a required argument? Or  does the position change make that not possible?

gabrieltseng: yea - it would screw up the indexing

gabrieltseng:
Added mean values. Specifically:

* **Mean monthly values**: these are (optionally) added to the `current` array
* **Mean yearly values**: these are (optionally) passed as an additional argument to the models. Different models deal with them differently.

src/models/data.py
```diff
@@ -184,15 +194,28 @@ def calculate_normalizing_array(self, data_vars: List[str]):
         for var in data_vars:
             for norm_var in normalizing_dict_keys:
                 if var.endswith(norm_var):
-                    mean.append(self.normalizing_dict[norm_var]['mean'])
-                    std.append(self.normalizing_dict[norm_var]['std'])
+                    var_mean = self.normalizing_dict[norm_var]['mean']
+                    var_std = self.normalizing_dict[norm_var]['std']
+                    if yearly_mean:
+                        var_mean = np.mean(var_mean)
```
tommylees112: this is definitely across all of the data and not the batches right?

tommylees112: how does this differ from the `yearly_mean` defined below on L230:

```python
        yearly_mean = x.mean(dim=['time', 'lat', 'lon'])
        yearly_mean_np = yearly_mean.to_array().values

        x = self._add_extra_dims(x, self.surrounding_pixels, self.monthly_means)


```

src/models/data.py
```diff
@@ -234,9 +263,11 @@ def ds_folder_to_np(self,
         y_np = y_np.reshape(y_np.shape[0], y_np.shape[1], y_np.shape[2] * y_np.shape[3])
         y_np = np.moveaxis(y_np, -1, 0).reshape(-1, 1)
 
-        if self.normalizing_array is not None:
+        if (self.normalizing_array is not None) and (self.normalizing_array_ym is not None):
             x_np = (x_np - self.normalizing_array['mean']) / (self.normalizing_array['std'])
             train_latlons = train_latlons / [90, 180]
+            yearly_mean_np = (yearly_mean_np - self.normalizing_array_ym['mean']) / \
```
tommylees112: so we're not including the yearly mean data we are also normalizing the data by the annual mean?

Does that make sense to duplicate that information? Wouldn't normalisation by the global mean and then including the annual mean have all of that information already incorporated?

gabrieltseng: `normalizing_array_ym` is the global mean information, fitted to an array that has the same dimensions as the yearly mean.

So normalization only happens using the global mean

src/models/data.py
```diff
@@ -297,24 +332,33 @@ def ds_folder_to_np(self,
...
-    def _add_surrounding(x: xr.Dataset, surrounding_pixels: int) -> xr.Dataset:
-
-        lat_shifts = lon_shifts = range(-surrounding_pixels, surrounding_pixels + 1)
-        for var in x.data_vars:
-            for lat_shift in lat_shifts:
-                for lon_shift in lon_shifts:
-                    if lat_shift == lon_shift == 0: continue
-                    shifted_varname = f'lat_{lat_shift}_lon_{lon_shift}_{var}'
-                    x[shifted_varname] = x[var].shift(lat=lat_shift, lon=lon_shift)
+    def _add_extra_dims(x: xr.Dataset, surrounding_pixels: Optional[int],
```
tommylees112: This is a great idea. Helps to clean some of the code up

src/models/data.py
gabrieltseng: What happens on line 200 is calculation of the `normalizing array` for the `yearly_mean_array` calculated in line 230

tests/models/test_data.py
```diff
@@ -196,27 +197,47 @@ def __init__(self):
...
-                        org = x_with_more.VHI.isel(time=0, lon=5, lat=5).values
-                        shifted = x_with_more[shifted_var_name].isel(time=0,
-                                                                     lon=5 + lon,
-                                                                     lat=5 + lat).values
-                        assert org == shifted, f"Shifted values don't match!"
+            if surrounding_pixels is not None:
+                for lat in [-1, 0, 1]:
+                    for lon in [-1, 0, 1]:
+                        if lat == lon == 0:
+                            assert f'lat_{lat}_lon_{lon}_{data_var}' not in shifted_mean_vars, \
```
tommylees112: What does this test for sorry?
* that the surrounding pixels are only included in the historical / `shifted` (shifted in time?) variables

gabrieltseng: This tests that `lat_1_lon_1_{var}` is equal to the variable, shifted by one latitude and longitude

src/models/regression.py
```diff
@@ -270,3 +215,32 @@ def _calculate_big_mean(self) -> np.ndarray:
             mean * size / total_size for mean, size in zip(means, sizes)
         ]
         return sum(weighted_means)
+
+    def _concatenate_data(self, x: Union[Tuple[Optional[np.ndarray], ...],
```
tommylees112: thanks this is a super useful function...

tests/models/neural_networks/test_ealstm.py
```diff
@@ -0,0 +1,129 @@
...
+    def test_save(self, tmp_path, monkeypatch):
+
+        features_per_month = 5
+        dense_features = [10]
+        hidden_size = 128
+        rnn_dropout = 0.25
+        include_latlons = True
+        include_pred_month = True
+        include_yearly_means = True
+        yearly_mean_size = 3
```
tommylees112: Sorry can you remind me why this has a size? Surely the yearly mean is a product of the number of `features_per_month` ?

gabrieltseng: Yes, but its not trivial to work out what the relationship is (for instance, if there are monthly aggregates or surrounding pixels). So its easier to pass this as an argument to the model directly

gabrieltseng:
Added standard deviations too

tests/models/neural_networks/test_ealstm.py
```diff
@@ -0,0 +1,129 @@
...
+        model = EARecurrentNetwork(hidden_size=hidden_size, dense_features=dense_features,
+                                   rnn_dropout=rnn_dropout, data_folder=tmp_path)
+        model.train()
+
+        captured = capsys.readouterr()
+        expected_stdout = 'Epoch 1, train smooth L1: 0.'
+        assert expected_stdout in captured.out
+
+        assert type(model.model) == EALSTM, \
+            f'Model attribute not an RNN!'
```
tommylees112: tiny change sorry - EALSTM ? or meant to be RNN?

src/models/neural_networks/ealstm.py
```diff
@@ -0,0 +1,313 @@
...
+        x_d : torch.Tensor
+            Tensor, containing a batch of sequences of the dynamic features. Shape has to match
+            the format specified with batch_first.
+        x_s : torch.Tensor
+            Tensor, containing a batch of static features.
+        Returns
+        -------
+        h_n : torch.Tensor
+            The hidden states of each time step of each sample in the batch.
+        c_n : torch.Tensor]
```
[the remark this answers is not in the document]
gabrieltseng: Ah that looks like a typo - I just copied this from the ealstm repo

src/models/neural_networks/ealstm.py
```diff
@@ -0,0 +1,313 @@
...
+            The cell states of each time step of each sample in the batch.
+        """
+        if self.batch_first:
+            x_d = x_d.transpose(0, 1)
+
+        seq_len, batch_size, _ = x_d.size()
+
+        h_0 = x_d.data.new(batch_size, self.hidden_size).zero_()
```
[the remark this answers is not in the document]
gabrieltseng: This is making a tensor of all zeros:
https://pytorch.org/docs/stable/tensors.html?highlight=zero_#torch.Tensor.zero_

tests/models/test_data.py
```diff
@@ -197,11 +197,13 @@ def __init__(self):
             assert np.all(x_train_data.current == expected), f"Expected to " \
                 "find the target_time data for the non target variables"
 
-        assert x_train_data.yearly_means.shape[1] == 4
+        assert x_train_data.yearly_aggs.shape[1] == 4 * 2  # means and std
```
tommylees112: 4 is the number of variables yes?

src/models/neural_networks/ealstm.py
```diff
@@ -0,0 +1,313 @@
...
+        # calculate input gate only once because inputs are static
+        bias_s_batch = (self.bias_s.unsqueeze(0).expand(batch_size, *self.bias_s.size()))
+        i = torch.sigmoid(torch.addmm(bias_s_batch, x_s, self.weight_sh))
+
+        # perform forward steps over input sequence
+        for t in range(seq_len):
+            h_0, c_0 = h_x
+
+            # calculate gates
+            gates = (torch.addmm(bias_batch, h_0, self.weight_hh) +
```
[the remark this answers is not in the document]
gabrieltseng: `i` is the input gate, and `g[t]` is the candidate cell input at timestep `t`. It goes "through" the input gate `i` (which you've correctly identified is simply element wise multiplication) before being added to the cell state.

tests/models/test_data.py
gabrieltseng: Yes

tests/models/neural_networks/test_ealstm.py
gabrieltseng: Thank you! fixed

src/models/data.py
```diff
@@ -230,15 +230,20 @@ def ds_folder_to_np(self,
         yearly_mean = x.mean(dim=['time', 'lat', 'lon'])
         yearly_mean_np = yearly_mean.to_array().values
 
+        yearly_std = x.std(dim=['time', 'lat', 'lon'])
+        yearly_std_np = yearly_std.to_array().values
+
+        yearly_agg = np.concatenate((yearly_mean_np, yearly_std_np))
```
tommylees112: what are the dimensions of this array then?

src/models/neural_networks/ealstm.py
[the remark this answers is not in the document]
gabrieltseng: fixed!

src/models/data.py
gabrieltseng: `yearly_agg.shape == (num_variables * 2,)`

src/models/data.py
tommylees112: gotcha!

PR #79, opened 2019-07-29
[minor] Remove standard deviation information from the aggregations

gabrieltseng:
As described. Experiments show they decrease model performance.


gabrieltseng:
It's easier to start from scratch than to deal with the conflicts

PR #80, opened 2019-07-30
Exporter for SRTM topography

gabrieltseng:
An exporter for SRTM topography data.

Mostly makes use of the [elevation package](http://elevation.bopen.eu/en/stable/) to download the data (which is why tests are a little sparse - the package has its own tests).


src/exporters/srtm.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+
+    dataset = 'srtm'
+
+    @staticmethod
+    def _region_to_tuple(region: Region) -> Tuple[float, float, float, float]:
+        return region.lonmin, region.latmin, region.lonmax, region.latmax
+
+    @staticmethod
+    def _tiff_to_nc(tif_file: Path, nc_file: Path) -> None:
```
tommylees112: WOHOOOO this is perfect because working with `tiff` files can be a pain in the booty

tommylees112: any reason why `tiff` vs. `tif`?

src/exporters/srtm.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+
+        region = region_lookup[region_name]
+
+        output_tif = self.output_folder / f'{region_name}.tif'
+
+        if not output_tif.exists():
+            print(f'Downloading tiles. Saving as tif to {output_tif}')
+            try:
+                elevation.clip(bounds=self._region_to_tuple(region),
```
tommylees112: yeah using this package def makes it more concise! great work Gabi

src/exporters/srtm.py
gabrieltseng: no haha - arbitrary
I guess it would be more consistent to have either `tif_to_nc` or `tiff_to_netcdf` - I'll make that quick change

PR #82, opened 2019-07-31
Preprocessor and engineer for static variables

gabrieltseng:
Incorporates static variables (specifically landcover type and topography) into the models.

The static data is stored in `data/features/static` and is a new attribute in the training data dataclass.

Requires CDO for the SRTM preprocessor.

### DIFF
--- a/src/exporters/srtm.py
+++ b/src/exporters/srtm.py
@@ -70,7 +70,8 @@ class SRTMExporter(BaseExporter):
                 elevation.clip(bounds=self._region_to_tuple(region),  # type: ignore
                                output=output_tif.resolve().as_posix(),
                                product=product,
-                               max_download_tiles=max_download_tiles)
+                               max_download_tiles=max_download_tiles,
+                               margin='1')
             except Exception as e:
                 print(e)
 
--- a/src/models/data.py
+++ b/src/models/data.py
@@ -329,7 +329,7 @@ class _BaseIter:
     def _calculate_static(self, num_instances: int) -> np.ndarray:
         static_np = self.static.to_array().values  # type: ignore
         static_np = static_np.reshape(static_np.shape[0], static_np.shape[1] * static_np.shape[2])
-        static_np = np.moveaxis(static_np, -1, 0).reshape(-1, 1)
+        static_np = np.moveaxis(static_np, -1, 0)
         assert static_np.shape[0] == num_instances
 
         if self.static_normalizing_dict is not None:
--- a/src/models/data.py
+++ b/src/models/data.py
@@ -20,6 +20,7 @@ class TrainData:
     # normalized
     latlons: Union[np.ndarray, torch.Tensor]
     yearly_aggs: Union[np.ndarray, torch.Tensor]
+    static: Union[np.ndarray, torch.Tensor, None]
 
 
 @dataclass
@@ -88,6 +89,8 @@ class DataLoader:
         Whether to include the monthly aggregates (mean and std across all spatial values) for
         the input variables. These will be additional dimensions to the historical
         (and optionally current) arrays
+    static: bool = True
+        Whether to include static data
     """
     def __init__(self, data_path: Path = Path('data'), batch_file_size: int = 1,
                  mode: str = 'train', shuffle_data: bool = True,
@@ -97,7 +100,8 @@ class DataLoader:
                  pred_months: Optional[List[int]] = None,
                  to_tensor: bool = False,
                  surrounding_pixels: Optional[int] = None,
-                 monthly_aggs: bool = True) -> None:
+                 monthly_aggs: bool = True,
+                 static: bool = False) -> None:
 
         self.batch_file_size = batch_file_size
         self.mode = mode
@@ -119,6 +123,18 @@ class DataLoader:
         self.surrounding_pixels = surrounding_pixels
         self.monthly_aggs = monthly_aggs
         self.to_tensor = to_tensor
+
+        self.static = None
+        self.static_normalizing_dict = None
+
+        if static:
+            static_data_path = data_path / 'features/static/data.nc'
+            self.static = xr.open_dataset(static_data_path)
+
+            if normalize:
+                static_normalizer_path = data_path / 'features/static/normalizing_dict.pkl'
+                with static_normalizer_path.open('rb') as f:
+                    self.static_normalizing_dict = pickle.load(f)
 
     def __iter__(self):
         if self.mode == 'train':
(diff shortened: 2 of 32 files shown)


tests/preprocess/test_esa_cci.py
```diff
@@ -105,33 +110,28 @@ def test_preprocess(self, tmp_path, cleanup):
...
+            tmp_path / 'interim/static/esa_cci_landcover_'
+            'preprocessed' / 'esa_cci_landcover_kenya_one_hot.nc'
         )
         assert expected_out_processed.exists(), \
             'expected a processed folder'
 
         # check the subsetting happened correctly
         out_data = xr.open_dataset(expected_out_processed)
-        expected_dims = ['lat', 'lon', 'time']
+        expected_dims = ['lat', 'lon']
```
tommylees112: why no `time` dimension? don't we need this to concatenate all files? even if we are making them `static`?

tests/exporters/test_cds.py
```diff
@@ -65,6 +68,7 @@ def test_area(self):
         expected_str = '6.002/33.501/-5.202/42.283'
         assert kenya_str == expected_str, f'Got {kenya_str}, expected {expected_str}!'
 
+    @pytest.mark.xfail(reason='cdsapi may not be installed')
```
tommylees112: this is bloody handy haha

src/preprocess/srtm.py
```diff
@@ -0,0 +1,78 @@
...
+from shutil import rmtree
+import xarray as xr
+import os
+
+from typing import Optional
+
+from .base import BasePreProcessor
+
+
+class SRTMPreprocessor(BasePreProcessor):
```
tommylees112: amazing love it

src/preprocess/srtm.py
```diff
@@ -0,0 +1,78 @@
...
+    dataset = 'srtm'
+    static = True
+
+    def regrid(self, ds: xr.Dataset, regrid: Path, method: str = 'remapbil') -> xr.Dataset:
```
tommylees112: are we happy with bilinear? i guess it shouldn' matter too much because we're going HIGH -> LOWER resolution. I only worry about it when we're making variance up using the bilinear methods from low->high resolution to be fair

src/preprocess/esa_cci.py
```diff
@@ -35,43 +35,20 @@ def create_filename(self, netcdf_filepath: str,
...
-        # perform the remap
-        for k, v in remap.items(): new_array[array == k] = v
-
-        # reassign values to new ds
-        new_ds['lc_class'] = (['time', 'lat', 'lon'], new_array)
-
-        return new_ds
+        for idx, row in legend.iterrows():
+            value, label = row.code, row.label
+            ds[f'{label}_one_hot'] = ds.lc_class.where(ds.lc_class == value, 0).clip(min=0, max=1)
```
tommylees112: ooo nice use of `.clip`

tommylees112: if `not one_hot_encode` do we use the old values? The remapping was mostly for plotting purposes

src/preprocess/esa_cci.py
```diff
@@ -183,20 +145,25 @@ class from the whole data range (1993-2019).
...
-                nc_files)
-            print("\nOutputs (errors):\n\t", outputs)
-
-        self.merge_files(subset_str, resample_time, upsampling)
+        for file in nc_files:
+            self._preprocess_single(file, subset_str, regrid)
+
+        ds = xr.open_mfdataset(self.get_filepaths('interim'))
+
+        ds = get_modal_value_across_time(ds.lc_class).to_dataset()
```
tommylees112: so we're having one landcover across all timesteps! I can sign up to this

src/models/neural_networks/base.py
```diff
@@ -146,7 +151,8 @@ def train(self, num_epochs: int = 1,
                                                 self._one_hot_months(x[1]),
                                                 x[2],
                                                 x[3],
-                                                x[4])
+                                                x[4],
+                                                x[5])
```
tommylees112: hehe how many of these can we get in ? ;)

src/models/neural_networks/ealstm.py
```diff
@@ -103,32 +115,38 @@ def _initialize_model(self, x_ref: Optional[Tuple[torch.Tensor, ...]]) -> nn.Mod
...
         self.include_pred_month = include_pred_month
         self.include_latlons = include_latlons
         self.include_yearly_agg = False
+        self.include_static = False
 
-        assert include_latlons or (yearly_agg_size is not None), \
-            "Need one of latlons or yearly mean for the static input"
-        static_size = 0
+        assert include_latlons or (yearly_agg_size is not None) or (static_size is not None), \
+            "Need at least one of {latlons, yearly mean, static} for the static input"
```
tommylees112: so are `latlons // yearly_mean` not percieved as `static` yet because they are derivatives of the dynamic variables? Would it make sense to save a `static` file of these into the `interim/static` folder so its absolutely clear the data being used? Or is that overkill and wasting storage?

src/exporters/srtm.py
```diff
@@ -70,7 +70,8 @@ def export(self, region_name: str = 'kenya',
                 elevation.clip(bounds=self._region_to_tuple(region),  # type: ignore
                                output=output_tif.resolve().as_posix(),
                                product=product,
-                               max_download_tiles=max_download_tiles)
+                               max_download_tiles=max_download_tiles,
+                               margin='1')
```
tommylees112: what does `mrgin` do?

src/models/data.py
```diff
@@ -124,6 +128,18 @@ def __init__(self, data_path: Path = Path('data'), batch_file_size: int = 1,
         self.to_tensor = to_tensor
         self.ignore_vars = ignore_vars
 
+        self.static = None
+        self.static_normalizing_dict = None
+
+        if static:
+            static_data_path = data_path / 'features/static/data.nc'
```
tommylees112: so static is always written to one `data.nc` file?

src/models/neural_networks/ealstm.py
gabrieltseng: Yea - because they are calculated from the dynamic variables (just before begin used), it would be hard to save them in the static folder

I think if we document the dataloader well it should be sufficient?

src/preprocess/esa_cci.py
gabrieltseng: yea - so not at the moment. I though it would actually be easier to go from the one hot encoding to a remapped value.

Should I preserve the remapping function?

tests/preprocess/test_esa_cci.py
gabrieltseng: (see below)

src/models/data.py
gabrieltseng: yup!

src/exporters/srtm.py
gabrieltseng: so we define the region to download. The margin downloads an extra `margin` of lat / lon to download.

PR #83, opened 2019-08-02
Clean up and document scripts

gabrieltseng:
1. Clean up the scripts a little.
2. Add a README in the scripts explaining what each of the scripts do
3. Update the main README (comment out the bits about the run.py file, and point to the scripts as the entrypoints to the pipeline instead)


scripts/export.py
```diff
@@ -85,10 +85,48 @@ def export_esa():
...
+        max_year=max_year,
+        min_month=min_month,
+        max_month=max_month,
+        max_leadtime=max_leadtime,
+        pressure_levels=pressure_levels,
+        n_parallel_requests=n_parallel_requests,
+    )
+
+
 if __name__ == '__main__':
```
tommylees112: shall we change this to only download a few years of data?

gabrieltseng: hmm - I wanted to keep it the same as how we are using it
sort of as a 'defaults argument' list, until we update the pipeline configuratons

PR #84, opened 2019-08-02
Docs/notebooks

tommylees112:
writing notebooks to `notebooks/docs` to introduce users to the pipeline


gabrieltseng:
Merged so that the documentation is available from `master`

PR #85, opened 2019-08-02
Remove standard deviation info

gabrieltseng:
As in https://github.com/esowc/ml_drought/pull/79

Also fixes a bug in how nan indices are calculated in the nowcaster experiment.

### DIFF
--- a/src/models/data.py
+++ b/src/models/data.py
@@ -412,7 +412,7 @@ class _BaseIter:
                         historical_nans_summed == 0
                     ) & (y_nans_summed == 0) & (
                         current_nans_summed == 0
-                    )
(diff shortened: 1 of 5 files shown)


src/models/data.py
```diff
@@ -417,7 +412,7 @@ def ds_folder_to_np(self,
                         historical_nans_summed == 0
                     ) & (y_nans_summed == 0) & (
                         current_nans_summed == 0
-                    )
+                    ) & (static_nans_summed == 0)
```
tommylees112: ah was this the offending line?

gabrieltseng: yes! oops

PR #86, opened 2019-08-02
Scripts/end to end

tommylees112:


scripts/draft/run_end_to_end.py
```diff
@@ -0,0 +1,179 @@
...
+    # preprocess
+    print("** Running Preprocessors **")
+    preprocess_data(data_path)
+
+    # engineer
+    print("** Running Engineer **")
+    engineer_data(data_path)
+
+    # models
+    print("** Running Models **")
```
gabrieltseng: Missing the models() function!

src/engineer/base.py
```diff
@@ -54,7 +54,8 @@ def _process_static(self):
         # lat and lon
 
         output_file = self.static_output_folder / 'data.nc'
-        assert not output_file.exists(), 'A static data file already exists!'
+        if not output_file.exists():
+            warnings.warn('A static data file already exists!')
```
gabrieltseng: This should be `if output_file.exists()` (it's currently `if not`).

Also, after the warning perhaps there could be a `return None` to make sure the static stuff isn't run (since it won't work)

tommylees112: Good point thankyou

PR #88, opened 2019-08-05
Calculate normalization values uniformly across time

gabrieltseng:
As described. We previously calculated normalization values independently for each time step - this really doesn't make much sense, since the model is trying to learn patterns of variations across time.

### DIFF
--- a/src/models/data.py
+++ b/src/models/data.py
@@ -194,6 +194,8 @@ class _BaseIter:
         self.static = loader.static
         self.static_normalizing_dict = loader.static_normalizing_dict
 
+        self.static_array: Optional[np.ndarray] = None
+
         if self.shuffle:
             # makes sure they are shuffled every epoch
             shuffle(self.data_files)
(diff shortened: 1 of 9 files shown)


tests/models/test_regression.py
```diff
@@ -61,12 +61,9 @@ def test_train(self, tmp_path, capsys, use_pred_months, experiment, monthly_agg)
         x_add2, _, _ = _make_dataset(size=(5, 5), const=True, variable_name='temp')
         x = xr.merge([x, x_add1, x_add2])
 
-        norm_dict = {'VHI': {'mean': np.zeros((1, x.to_array().values.shape[1])),
-                             'std': np.ones((1, x.to_array().values.shape[1]))},
-                     'precip': {'mean': np.zeros((1, x.to_array().values.shape[1])),
-                                'std': np.ones((1, x.to_array().values.shape[1]))},
-                     'temp': {'mean': np.zeros((1, x.to_array().values.shape[1])),
-                              'std': np.ones((1, x.to_array().values.shape[1]))}}
+        norm_dict = {'VHI': {'mean': 0, 'std': 1},
```
tommylees112: what happens to the other variables?

src/models/data.py
```diff
@@ -269,7 +259,7 @@ def _calculate_aggs(self, x: xr.Dataset) -> np.ndarray:
 
         if (self.normalizing_dict is not None) and (self.normalizing_array is None):
             self.normalizing_array_ym = self.calculate_normalizing_array(
-                list(yearly_mean.data_vars), yearly_agg=True)
```
tommylees112: we're removing the `yearly_agg` because we now capture that using the normalization defined across all timesteps fed into the model in the `X` data right?

tommylees112: also would `yearly_agg` make sense if we don't include 12 previous months? (say 3 prior months)

tests/models/test_regression.py
gabrieltseng: oops! this is a mistake.

Since they are not in the norm dict, they wouldn't be normalized, but I'll add them back to the test

src/models/data.py
gabrieltseng: We're removing the `yearly_agg` argument from the function, because now there doesn't need to be a distinction. However, the `yearly_agg` will still be there

tests/models/test_regression.py
gabrieltseng: What other variables do you mean?

gabrieltseng: whereas before, we had a vector of length 12 (for 12 timesteps) and we divided it by 12 unique values, now we divide all 12 values by the same value.

tommylees112: as in `temp` etc.

src/models/data.py
tommylees112: hmm so the yearly aggregation is still sent to the model even if we define our `X.nc`  in the engineer as 3 timesteps?

tests/models/test_regression.py
gabrieltseng: they are still there (lines 65, 66)

src/models/data.py
gabrieltseng: yea - yearly aggregation may be a misnomer, since the aggregation is calculated from the input data.

So if the engineer produces 3 timesteps, then the aggregation is accross 3 timesteps

tommylees112: cool!
merge this bad boy

PR #89, opened 2019-08-05
EALSTM composed of linear layers

gabrieltseng:
An identical implementation of what we had originally (see the test), but now using linear modules so that it can be piped through shap.

### DIFF
--- a/src/models/neural_networks/ealstm.py
+++ b/src/models/neural_networks/ealstm.py
@@ -223,60 +223,48 @@ class EALSTMCell(nn.Module):
         self.hidden_size = hidden_size
         self.batch_first = batch_first
 
-        # the h layers have no bias - since we add the outputs of these layers
-        # only one of them need a bias term
-
-        self.forget_gate_i = nn.Sequential(*[
-            nn.Linear(in_features=input_size_dyn, out_features=hidden_size,
-                      bias=True), nn.Sigmoid()])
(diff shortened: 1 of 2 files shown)


src/models/neural_networks/ealstm.py
```diff
@@ -209,10 +209,131 @@ def forward(self, x, pred_month=None, latlons=None, current=None, yearly_aggs=No
...
+        self.output_gate_h = nn.Linear(in_features=hidden_size, out_features=hidden_size,
+                                       bias=True)
+
+        self.sigmoid = nn.Sigmoid()
+        self.tanh = nn.Tanh()
+
+        self.reset_parameters()
+
+    def reset_parameters(self):
+        self._reset_i(self.forget_gate_i)
```
tommylees112: what is `reset_i`?

```python
    def _reset_i(layer):
        nn.init.orthogonal(layer.weight.data)
```

tommylees112: it's setting the `i` matrices to random values?

gabrieltseng: It initializes the weights of the neural networks. `reset_i` initializes the weights of layers on the input (`x[t]`) and `reset_h` initializes weights of layer on the hidden vector `h_0`

This serves the same function as `init_weights` in the linear network.

PR #90, opened 2019-08-05
Exporters/admin region

tommylees112:
Download the Kenyan administrative regions from https://data.humdata.org


src/exporters/admin_boundaries.py
```diff
@@ -0,0 +1,132 @@
...
+        print(f'Renamed file to: {new_fname.as_posix()}')
+
+
+class KenyaAdminExporter(OCHAExporter):
+    country = 'kenya'
+
+    def export(self) -> None:
+        """Export functionality for the Kenya Admin Boundaries as .shp files
+        """
+        urls = {
```
gabrieltseng: nit: Should these URLs be class attributes?

tests/exporters/test_srtm.py
```diff
@@ -1,4 +1,5 @@
 import unittest
+import unittest.mock
```
gabrieltseng: Is this import necessary?

tommylees112: it was on my machine!

src/exporters/admin_boundaries.py
tommylees112: so they can be accessed as `self.urls`?

tommylees112: so they can be accessed as `self.urls`?

if you think so! yeah for sure

tommylees112:
@gabrieltseng do you want this data to be stored elsewhere? It's in the raw folder which is only picked up by the preprocessors?

The preprocessors save the outputs to `data/analysis/boundaries` which is in the preprocessor PR which has been merged

src/exporters/admin_boundaries.py
tommylees112: okay fixed that!

PR #101, opened 2019-08-06
[minor] Add pred months to the static data in the EALSMT

gabrieltseng:
As described


gabrieltseng:
sort of. The static data isn't piped through the RNN, just concatenated the output hidden vector.

PR #102, opened 2019-08-07
Preprocess/admin region

tommylees112:
Preprocess the downloaded shapefile (`.shp`) into an xarray object with different values (like a categorical dataset) for different regions.

NOTE: Currently the regions have to be non-overlapping. I think if they do overlap it would just assign the later value. Is this a problem?

### DIFF
--- a/src/preprocess/base.py
+++ b/src/preprocess/base.py
@@ -28,6 +28,7 @@ class BasePreProcessor:
     """
     dataset: str
     static: bool = False
+    analysis: bool = False
 
     def __init__(self, data_folder: Path = Path('data')) -> None:
         self.data_folder = data_folder
@@ -43,7 +44,11 @@ class BasePreProcessor:
             else:
                 folder_prefix = self.dataset
 
-            self.out_dir = self.preprocessed_folder / f'{folder_prefix}_preprocessed'
+            if self.analysis:
+                self.out_dir = self.data_folder / 'analysis' / f'{self.dataset}_preprocessed'
+            else:
+                self.out_dir = self.preprocessed_folder / f'{folder_prefix}_preprocessed'
+
             if not self.out_dir.exists():
                 self.out_dir.mkdir(parents=True)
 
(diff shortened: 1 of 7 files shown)


src/preprocess/base.py
```diff
@@ -43,7 +44,11 @@ def __init__(self, data_folder: Path = Path('data')) -> None:
             else:
                 folder_prefix = self.dataset
 
-            self.out_dir = self.preprocessed_folder / f'{folder_prefix}_preprocessed'
+            if self.analysis:
```
tommylees112: wanted this functionality to remain with the base class to avoid writing the same function for the child class? does that make sense?

src/preprocess/utils.py
```diff
@@ -53,3 +62,136 @@ def select_bounding_box(ds: xr.Dataset,
             f"Sometimes this means that the latmin, latmax are the wrong way around. " \
             f"Try switching the order of latmin, latmax"
     return ds_slice
+
+
+class SHPtoXarray:
```
tommylees112: put this into a class so I can do the conditional imports 

```python
        global features
        if features is None:
            from rasterio import features

        global Affine
        if Affine is None:
            from affine import Affine

        global gpd
        if gpd is None:
            import geopandas as gpd

        global Polygon
        if Polygon is None:
            from shapely.geometry import Polygon

```

tests/utils.py
```diff
@@ -45,3 +48,29 @@ def _create_dummy_precip_data(tmp_path,
     precip.to_netcdf(data_dir / 'chirps_kenya.nc')
 
     return data_dir
+
+
+class CreateSHPFile:
```
tommylees112: again put into a class for the conditional imports

src/preprocess/base.py
gabrieltseng: yea, this makes sense to me

gabrieltseng:
I would imagine the overlap to be very small, no?

PR #103, opened 2019-08-07
Add variational dropout for recurrent models

gabrieltseng:
As in https://arxiv.org/pdf/1512.05287.pdf

The motivation is that dropout between hidden timesteps makes it hard for the models to learn patterns across time, since activations are randomly masked.

The solution is to apply the same dropout mask to each timestep in a minibatch.

### DIFF
--- a/src/models/neural_networks/utils.py
+++ b/src/models/neural_networks/utils.py
@@ -1,0 +1,25 @@
+import torch
+from torch import nn
+
+
+class VariationalDropout(nn.Module):
+    """
+    This ensures the same dropout is applied to each timestep,
+    as described in https://arxiv.org/pdf/1512.05287.pdf
+    """
+    def __init__(self, p):
+        super().__init__()
+
+        self.p = p
+        self.mask = None
+
+    def update_mask(self, x_shape, is_cuda):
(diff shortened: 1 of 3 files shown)


gabrieltseng:
Experiments with this are not so promising - chucking on a WIP to see if I can figure out why

tommylees112:
For reference (diagram from the paper):

Screenshot 2019-08-12 at 11 49 43

src/models/neural_networks/utils.py
```diff
@@ -0,0 +1,25 @@
...
+    as described in https://arxiv.org/pdf/1512.05287.pdf
+    """
+    def __init__(self, p):
+        super().__init__()
+
+        self.p = p
+        self.mask = None
+
+    def update_mask(self, x_shape, is_cuda):
+        mask = torch.bernoulli(torch.ones(x_shape) * (1 - self.p)) / (1 - self.p)
```
tommylees112: So it's randomly choosing parts of the input matrix (`x` here does not refer to the input data - `x` static/dynamic) in the layers and applying the SAME mask to all of those layers? 

e.g. on the RNN it might apply the same mask to all of the `input_gate` matrices?

gabrieltseng:
Closing, since we won't add it at the moment. Can be re-opened if necessary

PR #104, opened 2019-08-07
Update all the shap code (and other analysis)

gabrieltseng:
Feature importance analysis stalled out because of issues in the shap library.

These have now been fixed. This PR updates the code so that it works once again.

- [x] Update shap code for neural networks
- [x] Update shap code for regression
- [x] Notebook demonstrating everything
- [x] Move notebook functions into `src/analysis`


gabrieltseng:
Sorry for the big PR @tommylees112 ; effectively, what this PR does is make the `explain` functions a little easier to call.

The input arrays are massaged to actually work with shap, and some analysis code which (up to now) lived in notebooks has been added so that other notebooks can call them too.

src/models/data.py
```diff
@@ -32,6 +32,19 @@ class ModelArrays:
     latlons: Optional[np.ndarray] = None
 
 
+# The dict below maps the indices of the arrays returned
+# by the training iterator to their name. It should be updated
+# if new inputs are added
+idx_to_input = {
```
tommylees112: really good idea

src/models/regression.py
```diff
@@ -131,18 +132,25 @@ def explain(self, data: TrainData) -> Tuple[np.ndarray, ...]:
...
+            analysis_folder = self.model_dir / 'analysis'
+            if not analysis_folder.exists():
+                analysis_folder.mkdir()
 
-        # TODO: properly split the additional arrays
+            np.save(analysis_folder / f'shap_values.npy', explanations)
+            np.save(analysis_folder / f'input.npy', reshaped_x)
 
-        return historical.reshape(batch, timesteps, dims), additional
+        return explanations
```
tommylees112: thanks so much for doing all of this with the SHAP values

src/models/regression.py
```diff
@@ -131,18 +132,25 @@ def explain(self, data: TrainData) -> Tuple[np.ndarray, ...]:
...
 
-        historical = explanations[:, :timesteps * dims]
-        additional = explanations[:, timesteps * dims:]
+        if save_shap_values:
+            analysis_folder = self.model_dir / 'analysis'
+            if not analysis_folder.exists():
+                analysis_folder.mkdir()
 
-        # TODO: properly split the additional arrays
+            np.save(analysis_folder / f'shap_values.npy', explanations)
```
tommylees112: How do we connect which features correspond to which indices in the `.npy` arrays?

src/models/neural_networks/linear_network.py
```diff
@@ -29,6 +30,7 @@ def __init__(self, layer_sizes: Union[int, List[int]],
                          include_latlons, include_monthly_aggs, include_yearly_aggs,
                          surrounding_pixels, ignore_vars, include_static)
 
+        self.input_layer_sizes = copy(layer_sizes)
         if type(layer_sizes) is int:
```
tommylees112: Does this need to be done BEFORE L33 so that `input_layer_sizes` is also cast to a `List[int]` instead of being an `int`

gabrieltseng: The idea of `input_layer_sizes` is to exactly mirror what the input to the class was, so that it can easily be loaded again.

src/models/regression.py
gabrieltseng: aah... still a WIP!

PR #105, opened 2019-08-11
Analysis/admin region

tommylees112:
Create an `analyzer` object that calculates regional error statistics.

It calculates the error statistics for each model, and each region shapefile   

Also done:
- Add timestamp to `preds` so that we can check if times of `y.nc` and `preds_{yr}_{mth}.nc` are matching


src/models/base.py
```diff
@@ -138,15 +138,21 @@ def evaluate(self, save_results: bool = True,
                 json.dump(output_dict, outfile)
 
         if save_preds:
+            # TODO: assign time to preds
```
gabrieltseng: It looks like this has been done? Can we remove this comment?

src/analysis/region_analysis/region_analysis.py
```diff
@@ -0,0 +1,400 @@
...
+
+    def __init__(self,
+                 data_dir: Path = Path('data'),
+                 experiment: str = 'one_month_forecast',
+                 admin_boundaries: bool = True):
+        self.pred_variable: Optional[str] = None
+        self.true_variable: Optional[str] = None
+
+        self.data_dir: Path = data_dir
+        self.models_dir: Path = data_dir / 'models' / experiment
```
gabrieltseng: The structure of this directory is
`data_dir/models/experiment/<MODEL_NAME>/preds`

It probably makes sense to include a way to decide which model should be evaluated

src/analysis/region_analysis/region_analysis.py
```diff
@@ -0,0 +1,400 @@
...
+        ))
+        mae = mean_absolute_error(
+            group_model_performance.true_mean_value,
+            group_model_performance.predicted_mean_value
+        )
+        r2 = r2_score(
+            group_model_performance.true_mean_value,
+            group_model_performance.predicted_mean_value
+        )
+        return rmse, mae, r2
```
gabrieltseng: Can we save these too?
Since the class knows the structure of the `data_dir`, can it save these results to the `analysis` folder under `<experiment>_<model>.csv`or something

src/analysis/region_analysis/region_analysis.py
tommylees112:
```python
        if models is not None:
            self.models = self.models[np.isin(self.models, models)]
````

tommylees112: where `models` is provided as an initialisation argument

src/analysis/region_analysis/region_analysis.py
tommylees112: calculated for all models!

src/analysis/region_analysis/region_analysis.py
gabrieltseng: Ah I see - thanks :)

PR #106, opened 2019-08-12
[minor] Improvements to the models

gabrieltseng:
1. Introduces a 'months from pred' feature to the EALSTM and RNN. This is an additional dimension to the dynamic data, which tells the model which timestep it is looking at. 

2. Introduces instance-wise shuffling

### DIFF
--- a/src/models/data.py
+++ b/src/models/data.py
@@ -122,7 +122,7 @@ class DataLoader:
 
         self.batch_file_size = batch_file_size
         self.mode = mode
-        self.shuffle = shuffle_data
+        self.shuffle_data = shuffle_data
         self.clear_nans = clear_nans
         self.experiment = experiment
         self.data_files = self._load_datasets(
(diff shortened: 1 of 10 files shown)


src/models/data.py
```diff
@@ -298,6 +301,13 @@ def _calculate_historical(self, x: xr.Dataset,
                 list(x.data_vars))
             x_np = (x_np - self.normalizing_array['mean']) / (self.normalizing_array['std'])
 
+        if self.months_from_pred:
```
tommylees112: so this is the extra column which tells the model `which timestep it is looking at `?

what do you mean by timestep? An integer from 0-N for each month from 1981-2019?

That would account for a 'global warming' kinda signal to be fair

src/models/data.py
```diff
@@ -545,6 +558,17 @@ def __next__(self) -> Tuple[Tuple[Union[np.ndarray, torch.Tensor],
             final_x_ym = np.concatenate(out_x_ym, axis=0)
             final_y = np.concatenate(out_y, axis=0)
 
+            if self.shuffle_data:
```
tommylees112: can you talk me through the shuffling...

This is shuffling the years? So we see `1981-01` first but then `1999-05` next?

gabrieltseng:
Closing, since we won't add it at the moment. Can be re-opened if necessary

PR #109, opened 2019-08-13
Improved plotting of shap

gabrieltseng:
So far:

- Allow polished variable names to be plotted in the shap plots
- Be better about plotting the timestamp of the input data, instead of just the index

The results of the latest iteration of plots can be found at https://github.com/esowc/ml_drought/blob/analysis/plotting/notebooks/draft/15_gt_ealstm.ipynb


src/analysis/plot_shap.py
```diff
@@ -4,7 +4,23 @@
 
 import numpy as np
 
-from typing import Dict, List
+from typing import Dict, List, Tuple, Optional
+
+
+int2month = {
```
tommylees112: beautiful!

PR #111, opened 2019-08-14
Update VHI preprocessor to handle other variables as well

gabrieltseng:
As described.

Also, updates the test to require less memory.

Also, currently the output is `<DATASET>_preprocessed/<DATASET>_<SUBSET>.nc`, which is unnecessarily complicated. Replace it with `<DATASET>_preprocessed/data_<SUBSET>.nc`.

### DIFF
--- a/src/preprocess/vhi.py
+++ b/src/preprocess/vhi.py
@@ -66,7 +66,6 @@ class VHIPreprocessor(BasePreProcessor):
             netcdf_filepath,
             subset_name=subset_str,
         )
-        print(filename)
 
         # test if the file already exists
         if Path(f'{output_dir}/{filename}').exists():
--- a/src/preprocess/vhi.py
+++ b/src/preprocess/vhi.py
@@ -31,7 +31,7 @@ class VHIPreprocessor(BasePreProcessor):
 
     def __init__(self, data_folder: Path = Path('data'),
                  var: str = 'VHI') -> None:
-        assert var in ['VCI', 'VHI']
+        assert var in ['VCI', 'VHI', 'TCI']
         self.data_var = var
 
         super().__init__(data_folder, var)
--- a/src/preprocess/vhi.py
+++ b/src/preprocess/vhi.py
@@ -29,6 +29,9 @@ class VHIPreprocessor(BasePreProcessor):
 
     dataset = 'vhi'
 
+    raw_height: int = 3616
+    raw_width: int = 10000
+
     def __init__(self, data_folder: Path = Path('data'),
                  var: str = 'VHI') -> None:
         assert var in ['VCI', 'VHI', 'TCI']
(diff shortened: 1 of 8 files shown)


src/preprocess/vhi.py
```diff
@@ -242,13 +260,12 @@ def extract_timestamp(ds: Dataset,
         date = pd.to_datetime(atime)
         return date
 
-    @staticmethod
-    def create_lat_lon_vectors(ds: Dataset) -> Tuple[Any, Any]:
+    def create_lat_lon_vectors(self, ds: Dataset) -> Tuple[Any, Any]:
         """ read the `ds.attrs` and create new latitude, longitude vectors """
-        assert ds.WIDTH.size == 10000, \
+        assert ds.WIDTH.size == self.raw_width, \
```
tommylees112: Oh this is smart. thankyou

tests/preprocess/test_vhi.py
```diff
@@ -213,9 +213,66 @@ def test_alternative_region_interim_creation(self, tmp_path):
...
+
+    def test_VCI(self, tmp_path):
+        v = VHIPreprocessor(tmp_path, var='VCI')
+
+        # get filename
+        demo_raw_folder = (v.raw_folder / 'vhi' / '1981')
+        demo_raw_folder.mkdir(parents=True, exist_ok=True)
+        netcdf_filepath = demo_raw_folder / 'VHP.G04.C07.NC.P1981035.VH.nc'
+
+        raw_height, raw_width = 360, 100
```
tommylees112: is there not a `_make_dummy_vhi_dataset()` function? We could put all of these into a call to that funciton?

tommylees112:
```
    @staticmethod
    def _make_vhi_dataset():
        # build dummy .nc object
        height = list(range(0, 3616))
        width = list(range(0, 10000))
        vci = tci = vhi = np.random.randint(100, size=(3616, 10000))
        height = list(range(0, 3616 // 100))
        width = list(range(0, 10000 // 100))
        vci = tci = vhi = np.random.randint(100, size=(len(height), len(width)))

        ds = xr.Dataset(
            {'VCI': (['HEIGHT', 'WIDTH'], vci),
             'TCI': (['HEIGHT', 'WIDTH'], tci),
             'VHI': (['HEIGHT', 'WIDTH'], vhi)},
            coords={
                'HEIGHT': height,
                'WIDTH': width}
        )
        return ds

```

tests/preprocess/test_vhi.py
```diff
@@ -213,9 +213,66 @@ def test_alternative_region_interim_creation(self, tmp_path):
...
         STAR_VHP.G04.C07.NC_1981_8_31_ethiopia_VH.nc'.replace(' ', '')
         assert expected_out_path.exists(), \
             f'Expected processed file to be saved to {expected_out_path}'
         assert out == expected_out_path, f"Expected: {expected_out_path}, \
         Got: {out}"
+
+    def test_VCI(self, tmp_path):
+        v = VHIPreprocessor(tmp_path, var='VCI')
+
+        # get filename
```
tommylees112:
```python
raw_ds = self._make_vhi_dataset()
```

gabrieltseng: Thanks :)

tests/preprocess/test_vhi.py
gabrieltseng: yes! it was unused in all of the tests haha. Updated

PR #116, opened 2019-08-16
Allow neural nets to train on GPUs

gabrieltseng:


src/models/__init__.py
```diff
@@ -14,9 +14,11 @@
 
 
 def load_model(model_path: Path, data_path: Optional[Path] = None,
-               model_type: Optional[str] = None) -> Union[RecurrentNetwork,
-                                                          LinearNetwork,
-                                                          LinearRegression]:
+               model_type: Optional[str] = None,
+               device: Optional[str] = 'cpu') -> Union[RecurrentNetwork,
+                                                       LinearNetwork,
```
tommylees112: do we need an assert statement to ensure it's only `cpu` or `gpu`?

tests/models/neural_networks/test_rnn.py
```diff
@@ -43,11 +43,10 @@ def mocktrain(self):
         model.train()
         model.save_model()
 
-        assert (tmp_path / 'models/one_month_forecast/rnn/model.pkl').exists(), \
+        assert (tmp_path / 'models/one_month_forecast/rnn/model.pt').exists(), \
```
tommylees112: why are we changing to the pytorch save data file? is this a general fix or specific for GPUs?

src/models/neural_networks/base.py
```diff
@@ -233,8 +246,9 @@ def predict(self) -> Tuple[Dict[str, Dict[str, np.ndarray]], Dict[str, np.ndarra
                         val.x.historical, self._one_hot_months(val.x.pred_months),
                         val.x.latlons, val.x.current, val.x.yearly_aggs, val.x.static
                     )
-                    preds_dict[key] = preds.numpy()
-                    test_arrays_dict[key] = {'y': val.y.numpy(), 'latlons': val.latlons,
+                    preds_dict[key] = preds.cpu().numpy()
+                    test_arrays_dict[key] = {'y': val.y.cpu().numpy(),
```
tommylees112: are the tests always run on cpu?

tommylees112: sorry this is the base class

src/models/__init__.py
gabrieltseng: hmm if there are multiple gpus it could be a variety of devices (e.g. `cuda:1`).
PyTorch will throw up a fuss if the device is wrong, but it will know better than us.

tests/models/neural_networks/test_rnn.py
gabrieltseng: It's to ensure we can load it onto the correct device:
https://pytorch.org/tutorials/beginner/saving_loading_models.html#saving-loading-model-across-devices

gabrieltseng:
A neat side effect - removed lots of one time use packages, so that we could have a 'minimal environment' when training the models

PR #117, opened 2019-08-20
Calculate shap values for an entire month

gabrieltseng:
Calculate and save shap values for an entire month.

This allows the spatial behavior of shap values to be investigated.


src/analysis/plot_shap.py
```diff
@@ -119,3 +124,69 @@ def plot_shap_values(x: np.ndarray,
     plt.draw()
     if show:
         plt.show()
+
+
+def all_shap_for_file(test_file: Path,
```
tommylees112: sorry why is it called `all_shap_for_file` ? if we're producing spatial maps can we call the function something like:

`spatial_shap_for_dataset` or something that has the spatial component in it?

gabrieltseng: `all_shap_for_file` is because shap values are calculated for all the datapoints in the `test_file` that is passed - so its not just spatial shap, shap values for all the time points are calculated too.

To be honest, I hadn't given much thought to the function name though, so happy to change it if you have something else in mind.

tommylees112: okay that's fair ! sorry i misunderstood

PR #118, opened 2019-08-20
Allow static era5 variables to be processed too

gabrieltseng:
As described. Useful for soil type.

### DIFF
--- a/src/preprocess/era5.py
+++ b/src/preprocess/era5.py
@@ -4,14 +4,19 @@ import xarray as xr
 import multiprocessing
 from shutil import rmtree
 
-from typing import Optional
+from typing import Optional, List
 
 from .base import BasePreProcessor
+from ..utils import get_modal_value_across_time
 
 
 class ERA5MonthlyMeanPreprocessor(BasePreProcessor):
 
     dataset = 'reanalysis-era5-single-levels-monthly-means'
+
+    # some ERA5 variables need to be treated statically
+    # they are recorded here
+    static_vars = ['soil_type']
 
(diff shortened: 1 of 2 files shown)


src/preprocess/era5.py
```diff
@@ -50,6 +55,81 @@ def _preprocess_single(self, netcdf_filepath: Path,
...
+    def get_filepaths(self, folder: str = 'raw',
+                      filter_type: Optional[str] = None) -> List[Path]:
+        """
+        filter_type can be {None, 'static', 'dynamic'}
+        """
+        if folder == 'raw':
+            target_folder = self.raw_folder / self.dataset
+        else:
+            target_folder = self.interim
+        outfiles = self.filter_outfiles(list(target_folder.glob('**/*.nc')),
```
tommylees112: this is selecting the static variables to exclude right? like it

src/preprocess/era5.py
```diff
@@ -50,6 +55,81 @@ def _preprocess_single(self, netcdf_filepath: Path,
...
+                                        filter_type)
+        outfiles.sort()
+        return outfiles
+
+    def merge_files(self, subset_str: Optional[str] = 'kenya',
+                    resample_time: Optional[str] = 'M',
+                    upsampling: bool = False,
+                    filename: Optional[str] = None) -> None:
+
+        # first, static
```
tommylees112: `# first, dynamic`

gabrieltseng: Thanks Tommy; fixed!

PR #120, opened 2019-08-27
[WIP] Add an embedding layer for the months

gabrieltseng:
As per Yasser's comments, replaces the one hot encoding with an embedding.

~Currently, the size of the embedding is fixed at 12, but if this works well this could become an additional argument.~ The size of the embedding is an additional argument - if `None` is passed, the prediction month isn't passed to the model.

Also, removes the EALSTM cell written using `nn.Module`s, since the original one works fine with shap.


tommylees112:
@gabrieltseng is this still good to merge?

PR #121, opened 2019-08-28
Notebooks/analysis

tommylees112:
Producing the final notebooks as part of the ECMWF Summer of Weather Code.

Split into just two notebooks:
1) Pipeline.ipynb
- an overview of the pipeline and its functionality
- the diagrams used from before
- a relatively thorough overview of each component part of the pipeline
 
2) Analysis.ipynb
- the *play along* notebook that helps us to demonstrate the analysis that can be done with the pipeline results.

Requires: 
1. A zip file (`data/zip_data.zip`)
1. How to share this data? Can we use GoogleDrive? Too big too share via github unfortunately (~1GB)
1. We need the trained `model.pt` file (my one doesn't seem to produce our sota RMSE (i get an rmse ~40)
1. need help with the shap values. need to be computed so that can be plotted (using the static cell from @gabrieltseng's draft notebook


gabrieltseng:
Nice work Tommy.

Some of the notebooks seem incomplete / don't run? (Specifically, `18_tl_analysis_chirps_grid.ipynb`). It's probably fine, since its in the drafts, but is this a bug in code which should be fixed?

tommylees112:
okay i think i fixed that bug!

gabrieltseng:
Nice ! Could you rerun that notebook too (just prevents buggy notebooks from being uploaded)

gabrieltseng:
LGTM once that happens

PR #125, opened 2019-09-27
Add XGBoost

gabrieltseng:
@tommylees112 , as discussed.

This only really works on a big computer (haven't tested on the full dataset yet - we should do that on the ERA5 scale data, since it will be quite a bit smaller).

### DIFF
--- a/src/models/base.py
+++ b/src/models/base.py
@@ -160,6 +160,9 @@ class ModelBase:
 
     def _concatenate_data(self, x: Union[Tuple[Optional[np.ndarray], ...],
                                          TrainData]) -> np.ndarray:
+        """Takes a TrainData object, and flattens all the features the model
+        is using as predictors into a np.ndarray
+        """
 
         if type(x) is tuple:
             x_his, x_pm, x_latlons, x_cur, x_ym, x_static = x  # type: ignore
--- a/src/models/base.py
+++ b/src/models/base.py
@@ -4,7 +4,9 @@ import json
 import pandas as pd
 from sklearn.metrics import mean_squared_error
 
-from typing import cast, Any, Dict, List, Optional, Tuple
+from .data import TrainData
+
+from typing import cast, Any, Dict, List, Optional, Union, Tuple
 
 
 class ModelBase:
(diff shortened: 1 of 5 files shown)


src/models/base.py
```diff
@@ -155,3 +157,35 @@ def evaluate(self, save_results: bool = True,
                 ).set_index(['lat', 'lon', 'time']).to_xarray()
 
                 preds_xr.to_netcdf(self.model_dir / f'preds_{key}.nc')
+
+    def _concatenate_data(self, x: Union[Tuple[Optional[np.ndarray], ...],
+                                         TrainData]) -> np.ndarray:
+
```
tommylees112: can we add a docstring?

```
        """For loading ALL data from dataloaders into memory"""
```

src/models/gbdt.py
```diff
@@ -0,0 +1,202 @@
...
+from .base import ModelBase
+from .data import DataLoader, train_val_mask, TrainData
+
+xgb = None
+
+
+class GBDT(ModelBase):
+
+    """This only really works if you have a lot of memory
+    """
```
tommylees112:
```
    (all data has to be loaded into memory for the XGBRegressor)

```

tommylees112: just so it's clear why a user might need alot of memory 

(user being future us ... ;) )

src/models/gbdt.py
```diff
@@ -0,0 +1,202 @@
...
+            ignore_vars=self.ignore_vars, monthly_aggs=self.include_monthly_aggs,
+            static=self.include_static)
+
+        preds_dict: Dict[str, np.ndarray] = {}
+        test_arrays_dict: Dict[str, Dict[str, np.ndarray]] = {}
+
+        assert self.model is not None, 'Model must be trained!'
+
+        for dict in test_arrays_loader:
+            for key, val in dict.items():
```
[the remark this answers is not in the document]
tommylees112: scrap this i think your way of doing it is neater

src/models/gbdt.py
```diff
@@ -0,0 +1,202 @@
...
+        for x, y in train_dataloader:
+            input_train_x.append(self._concatenate_data(x))
+            input_train_y.append(y)
+
+        input_train_x_np = np.concatenate(input_train_x, axis=0)
+        input_train_y_np = np.concatenate(input_train_y)
+
+        fit_inputs = {'X': input_train_x_np, 'y': input_train_y_np}
+
+        if early_stopping is not None:
```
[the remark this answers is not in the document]
gabrieltseng: So this is an argument passed to the function. The names are a little confusing, but:

`early_stopping`: patience (number of epochs to wait without improvement). If its `None`, then early stopping isn't used.
`self.early_stopping` (i.e. the model attribute): whether or not to use early stopping

src/models/gbdt.py
```diff
@@ -0,0 +1,202 @@
...
+                 include_yearly_aggs: bool = True,
+                 surrounding_pixels: Optional[int] = None,
+                 ignore_vars: Optional[List[str]] = None,
+                 include_static: bool = True) -> None:
+        super().__init__(data_folder, batch_size, experiment, pred_months,
+                         include_pred_month, include_latlons, include_monthly_aggs,
+                         include_yearly_aggs, surrounding_pixels, ignore_vars,
+                         include_static)
+
+        self.early_stopping = False
```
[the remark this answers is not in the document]
gabrieltseng: see the comment above

src/models/base.py
gabrieltseng: Added!

src/models/gbdt.py
gabrieltseng: Added !

PR #126, opened 2019-10-11
Polish up the run.py stuff

gabrieltseng:
It works!

Try running it with the `minimal.json` file.

Next steps:
- [x] Add a `json` file for the experiments we have been running so far
- [x] Clean up the other configurations
- [x] Update readme


tommylees112:
this is HUUUUUGE for interacting with the pipeline :D :D

src/run.py
```diff
@@ -62,50 +68,103 @@ def export(self, export_args: Dict) -> None:
             'era5': ERA5Exporter,
             'vhi': VHIExporter,
             'era5POS': ERA5ExporterPOS,
+            'gleam': GLEAMExporter,
+            'chirps': CHIRPSExporter,
+            'srtm': SRTMExporter
```
tommylees112: can we add the KenyaAdminBoundaries exporter too? Or does this need a more involved function?

```
def run_region_exporter(selection):
    to_region = { 
        'kenya' : KenyaAdminExporter ,
        'ethiopia' : EthiopiaAdminExporter ,
        'east_africa' : EastAfricaAdminExporter
    }
```

Maybe I can do this in a separate PR?

tommylees112: probably makes most sense to do in a different PR

gabrieltseng: added, then removed.

On reflection, could we update the API for the boundaries exporting class - I think having a unique class for each export seems clunky, especially since (for the two which work, kenya and ethiopia), what happens in the export functions is essentially the same, just with a different url.

We could collapse it into a single class, or (like the engineer) have them be called from an overarching `Boundaries` class.

gabrieltseng: (my main concern is if we expand this to include all the countries, there will be an overwhelming amount of exporters / preprocessors for countries)

PR #134, opened 2019-11-02
Add black, ignore failing pep8

gabrieltseng:
Sorry about the huge diff, Tommy.

Probably the right approach here is to just take a general look and see if the format of the code works for you. If not, we can scrap this PR.

- [x] Makes sure all the code is black-compliant
- [x] Gets travis to make sure the code is black compliant
- [x] If PEP8 and black clash, black takes precedence. Gets flake8 to ignore the offending PEPs
- [x] Adds black to the environment


tommylees112:
I think `black` code looks good and really like the automatic changes. That being said i think we both want to be happy with the format?

I do think it adds another layer of help/guidance for writing good code right?

gabrieltseng:
yes! let's do it

PR #136, opened 2019-11-16
Add Morris explanations

gabrieltseng:
Also, adds a `get_dataloader` function to the model base class, which makes all the actual model code a little more readable - this is the majority of the diff (sorry!)


src/models/neural_networks/base.py
```diff
@@ -415,3 +352,49 @@ def make_shap_input(
         else:
             output_tensors.append(x.static[start_idx : start_idx + num_inputs])
         return output_tensors
+
+    def get_morris_gradient(self, x: TrainData) -> TrainData:
```
tommylees112: This is absolute class thanks @gabrieltseng !!!

PR #138, opened 2019-11-23
Allow Morris gradients to be saved

gabrieltseng:
Tommy, opened a PR so you can use the code.
The arrays are saved in the same way shap values are saved - this way, the utility functions we developed for shap values can also easily be used for Morris gradients.

A todo (~hopefully it can be included in this PR~ Done!) is to pass a TrainData object to the shap explainer too, and consolidate it into a single `explain` function, where the method (shap or morris) can be selected.


gabrieltseng:
@tommylees112 let me know if this works for you! if it does, I'll clean up the API

src/models/neural_networks/base.py
```diff
@@ -397,4 +406,20 @@ def get_morris_gradient(self, x: TrainData) -> TrainData:
             else:
                 output_dict[key] = None
 
+        if save_explanations:
```
tommylees112: is this in the original `TrainData` dimensions or is it a `latlon` grid like the original xarray data?

tommylees112: or have i misinterpreted how these gradients are calculated? Are they not also per pixel, per variable?

gabrieltseng: so this is the same as the shap values, in a numpy array. This means the `all_shap_for_file` function can be used (https://github.com/esowc/ml_drought/blob/master/src/analysis/plot_shap.py) to turn the outputs into a netcdf file.

tommylees112: the only issue i see with working with those `plot_shap` functions is the hardcoding of the months. Our data will very likely become daily / weekly at some point and we want the flexibility of working with hourly data too (e.g. if emulating ERA5 hourly). That's why i think writing conversions to xarray objects makes sense because we always have the timesteps/timestamps

gabrieltseng:
Merging to use some of the refactors to `data.py` seperately. I'll open a new PR for the npy to netcdf stuff

PR #140, opened 2019-12-10
Adds an unsupervised warm up for the models

gabrieltseng:
Inspired by [tile2vec](https://arxiv.org/pdf/1805.02855.pdf), pretrains the models by training the models to make embeddings that are far away from each other more different than embeddings that are close to one another.

It's a less rigid way of communicating the latlon information to the models, and decreases RMSE by ~1


tommylees112:
This is super cool Gabi! Thanks so much. Just reviewing now

src/models/neural_networks/triplet_data.py
```diff
@@ -0,0 +1,235 @@
...
+
+from ..data import DataLoader, _BaseIter, ModelArrays, TrainData
+
+from typing import Union, Tuple, Optional, List, Iterable
+
+
+class TripletLoader(DataLoader):
+    def __init__(
+        self,
+        neighbouring_distance: Union[float, Tuple[float, float]] = 2,
```
tommylees112: this is the `neighbourhood_size` parameter described in the tile2vec paper?

src/models/neural_networks/triplet_data.py
```diff
@@ -0,0 +1,235 @@
...
+    ) -> Tuple[TrainData, TrainData, TrainData]:
+        """
+        Given a single modelArray, returns a tuple containing
+        [AnchorData, NeighbourData, DistantData]
+        """
+        anchor_indices: List[int] = []
+        neighbour_indices: List[int] = []
+        distant_indices: List[int] = []
+
+        outer_distance = tuple(multiplier * val for val in distance)
```
tommylees112: whats the role of the multiplier?

src/models/neural_networks/ealstm.py
```diff
@@ -288,11 +289,14 @@ def forward(
 
         x = self.rnn_dropout(hidden_state[:, -1, :])
 
+        if return_embedding:
```
tommylees112: is this for interpreting the static embedding?

src/models/neural_networks/base.py
```diff
@@ -190,6 +199,139 @@ def predict(self) -> Tuple[Dict[str, Dict[str, np.ndarray]], Dict[str, np.ndarra
...
+                neighbouring_distance=neighbouring_distance,
+                multiplier=multiplier,
+                to_tensor=True,
+                shuffle_data=True,
+            )
+
+        # initialize the model
+        if self.model is None:
+            x_ref, _, _ = next(iter(train_dataloader))
+            model = self._initialize_model(self._input_to_tuple(x_ref))
```
tommylees112: does this train the LSTM model? Don't we need to initialise with a CNN as they use in Tile2Vec?

src/models/neural_networks/triplet_data.py
gabrieltseng: Yup!

src/models/neural_networks/ealstm.py
gabrieltseng: No - the loss in tile2vec compares the embedding, not the final value. This is to return that embedding for the loss, before it gets put through the final linear layer

src/models/neural_networks/triplet_data.py
gabrieltseng: It's basically to enforce a minimum distance between the neighbouring instance and the distant instance.

The neighbour will be within `neighbouring_distance` of the anchor. The distant instance will be further than `multiplier * neighbouring_distance` from the anchor

src/models/neural_networks/base.py
gabrieltseng: The principles of tile2vec can be used with any model that takes a raw input and outputs an embedding.

So yea, in this case it can also train the (EA)LSTM model

src/models/data.py
```diff
@@ -42,14 +42,33 @@ def concatenate(self, x: TrainData):
                     newval = torch.cat((val, getattr(x, key)), dim=0)
                 setattr(self, key, newval)
 
+    def __getitem__(self, index: Union[int, List[int]]) -> TrainData:
```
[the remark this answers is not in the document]
gabrieltseng: It's used here:

https://github.com/esowc/ml_drought/pull/140/files#diff-0326e9cb4ac2e09a4a57c4c5caa8c5a6R163

It basically allows the elements in `TrainData` to be indexed (so `train_data[0]` would give you a new `TrainData` instance, with the first element of all the attributes in `TrainData`.

src/models/neural_networks/triplet_data.py
tommylees112: gotcha! so basically enforcing how large an area our spatial differences should be over

src/models/neural_networks/ealstm.py
tommylees112: yes makes sense! Could this be used for interpreting the embedding layer too though?

src/models/neural_networks/base.py
tommylees112: okay gotcha.

So have i interpreted this correctly:

"We use the unsupervised learning algorithm described in Tile2Vec to pretrain (initialise) the weights of the EALSTM. This allows us to produce weights in the network that produce sensible spatial patterns. Mainly that pixels close together are more similar than pixels that are far apart."

src/models/neural_networks/triplet_data.py
tommylees112: have we already used that name or can we use it so that it's easier for people (future us) to go from paper to implementation?

src/models/neural_networks/base.py
gabrieltseng: yea, that's exactly right

src/models/neural_networks/ealstm.py
gabrieltseng: yea, 100%. Although here the "embedding" is the final output of the model before the linear regression layer

src/models/neural_networks/triplet_data.py
gabrieltseng: yup!

src/models/neural_networks/triplet_data.py
gabrieltseng: yea, I'll update this

gabrieltseng: updated!

PR #141, opened 2019-12-28
Model the delta (changes) rather than the raw values

tommylees112:
* add the `predict_delta` argument to models
* add the `ModelArray.to_xarray()` method for converting `ModelArray` objects to `xr.Dataset` objects


src/models/base.py
```diff
@@ -87,6 +94,18 @@ def __init__(
...
         self.device = "cpu"
 
+    def _convert_delta_to_raw_values(
+        self, x: xr.Dataset, y: xr.Dataset, y_var: str, order: int = 1
+    ) -> xr.Dataset:
+        """When calculating the derivative we need to convert the change/delta
+        to the raw value for our prediction.
+        """
+        # x.shape == (pixels, featurespreds)
+        prev_ts = x[y_var].isel(time=-order)
```
gabrieltseng: Should we take the negative of the passed order? Why not make it negative from the get go and pass the raw value?

src/models/base.py
```diff
@@ -174,6 +194,34 @@ def evaluate(self, save_results: bool = True, save_preds: bool = False) -> None:
                     .to_xarray()
                 )
 
+                if self.predict_delta:
+                    # get the NON-NORMALIZED data (ModelArrays)
+                    dl = self.get_dataloader(
```
gabrieltseng: Is is necessary to call a new dataloader here?

An alternative is to add an attribute to the `ModelArrays`, which (if `predict_delta == True`) contains the appropriate target xarray ( / numpy array - that might be easier).

That way, all of this can be abstracted away into the dataloader, and it should be a little cleaner since all the data will be there together.

src/models/data.py
```diff
@@ -63,11 +64,103 @@ def concatenate(self, x: ModelArrays) -> None:
         if type(self.y) is np.ndarray:
             self.y = np.concatenate((self.y, x.y), axis=0)
         else:
-            self.y = torch.cat((self.y, x.y,), dim=0)
+            self.y = torch.cat((self.y, x.y), dim=0)
 
         if self.latlons is not None:
             self.latlons = np.concatenate((self.latlons, x.latlons), axis=0)
 
+    def to_xarray(self) -> Tuple[xr.Dataset, xr.Dataset, Optional[xr.Dataset]]:
```
gabrieltseng: Nice - thanks for adding this

src/models/base.py
tommylees112: where would i pick up the data from the ModelArrays though? because aren't the `TestData` objects only opened in the `predict()` functions e.g. [here](https://github.com/esowc/ml_drought/blob/95b03918bbcfa751d32ee401b87c6535aae2b396/src/models/neural_networks/base.py#L168)

gabrieltseng: right - this would be in the dataloader step. That test arrays loader returns ModelArrays (https://github.com/esowc/ml_drought/blob/95b03918bbcfa751d32ee401b87c6535aae2b396/src/models/data.py#L662).

src/models/base.py
tommylees112: I think it is more interpretable remaining positive so that people know it is 'change since `n_timesteps` before'

gabrieltseng: yea, that makes sense!

src/models/base.py
tommylees112: Nice thanks for your help here Gabi! Much more sensible now I think (plus we're passing around numpy arrays instead of xarray objects)

PR #144, opened 2020-01-14
ASAL masks

gabrieltseng:
- [x] Add a preprocessor for the ASAL masks
- [x] Pass the masks to the dataloader
- [x] Use the masks to define the training subset
- [x] Get it working with the models


src/models/data.py
```diff
@@ -323,6 +330,18 @@ def __iter__(self):
     def __len__(self) -> int:
         return len(self.data_files) // self.batch_file_size
 
+    @staticmethod
+    def prepare_mask(spatial_mask: Optional[xr.DataArray]) -> Optional[xr.DataArray]:
```
tommylees112: do we definitely want the dataloader to be inverting the mask and not just use it 

`xr.Dataset.where(~mask)` ?

Leave this up to you i actually don't have a preference just interested in the reasoning for this inversion here!

src/models/data.py
```diff
@@ -541,6 +561,17 @@ def _calculate_static(self, num_instances: int) -> np.ndarray:
             self.static_array = static_np
         return self.static_array
 
+    def apply_spatial_mask(
```
tommylees112: can also use 

`xr.Dataset.where(mask)`

src/models/data.py
gabrieltseng: `where` is so much more elegant. I'll make that change.

In terms of the inversion, I was trying to think about which would be more intuitive - I figured since we call it a `mask`, then probably where `mask == 1` would be "masked out" (i.e. removed).

src/preprocess/base.py
```diff
@@ -159,10 +159,10 @@ def regrid(
             output_dict[var] = regridder(ds[var])
         ds = xr.Dataset(output_dict)
 
-        print(
-            f"Regridded from {(regridder.Ny_in, regridder.Nx_in)} "
-            f"to {(regridder.Ny_out, regridder.Nx_out)}"
-        )
+        # print(
+        #     f"Regridded from {(regridder.Ny_in, regridder.Nx_in)} "
```
tommylees112: is this remaining commented out?

src/preprocess/utils.py
```diff
@@ -212,7 +212,7 @@ def shapefile_to_xarray(
             if (values is not None) & (keys is not None)
         ]
         values = [
-            key
+            value
```
tommylees112: great catch!

src/models/data.py
gabrieltseng: so much better. Updated

src/preprocess/base.py
gabrieltseng: oops... fixed! thanks for catching this. Fixed!

src/models/data.py
gabrieltseng: Updated!

PR #149, opened 2020-01-25
Model improvements

gabrieltseng:
- [x] Pass the previous year's VCI (for the same target month) as static input to the model
- [x] Allow the model to predict normalized values, so it doesn't need to learn mean and std

### DIFF
--- a/src/models/data.py
+++ b/src/models/data.py
@@ -871,6 +871,7 @@ class _TrainIter(_BaseIter):
                         global_modelarrays.x.current,
                         global_modelarrays.x.yearly_aggs,
                         global_modelarrays.x.static,
+                        global_modelarrays.x.prev_y_var,
                     ),
                     global_modelarrays.y,
                 )
--- a/src/models/data.py
+++ b/src/models/data.py
@@ -24,6 +24,7 @@ class TrainData:
     latlons: Union[np.ndarray, torch.Tensor]
     yearly_aggs: Union[np.ndarray, torch.Tensor]
     static: Union[np.ndarray, torch.Tensor, None]
+    prev_y_var: Union[np.ndarray, torch.Tensor]
 
     def to_tensor(self, device: torch.device) -> None:
         for key, val in self.__dict__.items():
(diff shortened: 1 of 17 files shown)


src/models/data.py
```diff
@@ -42,6 +43,16 @@ def concatenate(self, x: TrainData):
                     newval = torch.cat((val, getattr(x, key)), dim=0)
                 setattr(self, key, newval)
 
+    def filter(self, filter_array: Union[np.ndarray, torch.Tensor]) -> None:
+        # noqa because we just want to make sure these to have the same type,
```
tommylees112: what's noqa?

src/models/data.py
```diff
@@ -42,6 +43,16 @@ def concatenate(self, x: TrainData):
                     newval = torch.cat((val, getattr(x, key)), dim=0)
                 setattr(self, key, newval)
 
+    def filter(self, filter_array: Union[np.ndarray, torch.Tensor]) -> None:
```
tommylees112: What is this function doing?

src/models/data.py
```diff
@@ -420,6 +443,33 @@ def __init__(self, loader: DataLoader) -> None:
     def __iter__(self):
         return self
 
+    def _get_prev_y_var(
```
tommylees112: is there any way we can pass in the mean of all the previous months? So in the train data this will be either 1 (for 1982) or 29 for 2010 (29 previous years of that month).

In the test it would just be the mean of the training data (like a climatology)

tommylees112: maybe for a future PR?

tommylees112: maybe that should be `get_mean_y_value` or something

src/models/data.py
gabrieltseng: Tells flake8 to ignore the line

src/models/data.py
gabrieltseng: Before, we had a bunch of lines which were
`historical = historical[notnan_indices]`

now, we can filter the entire `TrainData` object, so

`traindata.filter(notnan_indices)`

src/models/data.py
gabrieltseng: yea. We sort of have that with the yearly_aggs, but it could definitely be revisited.

src/models/data.py
tommylees112: gotcha

PR #152, opened 2020-01-31
Engineer/test similar arrays

tommylees112:
added the line to the engineer comparison:

`np.sort(datasets[idx][dim].values), np.sort(coords[dim])`

because the `lat`/`lon` coords aren't always sorted
and xarray knows how to deal with them anyway


tommylees112:
Merging this badboy!

PR #157, opened 2020-03-04
Static/static data

tommylees112:
Convert the static data -> temporally static (except for monthly climatology) ...

Now these are fixed!

Key changes to check:
- src/engineer/base.py
- src/models/data.py
- src/models/neural_networks/ealstm.py --> # append prev_y to DYNAMIC data


src/engineer/base.py
```diff
@@ -47,9 +48,42 @@ def engineer(
...
+        test_year = [test_year] if not isinstance(test_year, Iterable) else test_year
+        dynamic_ds = dynamic_ds.sel(time=slice(min_year, str(min(test_year) - 1)))
+        assert all(
+            int(yr) not in np.unique(dynamic_ds["time.year"].values) for yr in test_year
+        )
+
+        ones = xr.ones_like(static_ds)
+        ones_da = ones[[v for v in ones.data_vars][0]]
+
+        # 1. create global means ds
```
tommylees112: This is where I add the `Global_Means` and the `Pixel_Means`

src/models/data.py
```diff
@@ -694,14 +721,17 @@ def ds_folder_to_np(
 
         x, y = self.apply_spatial_mask(x, y)
 
-        yearly_agg = self._calculate_aggs(
-            x
-        )  # before to avoid aggs from surrounding pixels
+        if self.incl_yearly_aggs:
+            assert False
```
tommylees112: just put this in because i wanted to check that the flag was working - REMOVE BEFORE END

src/engineer/base.py
tommylees112: these should be constant across each timestep

src/engineer/base.py
```diff
@@ -47,10 +48,49 @@ def engineer(
 
         self._process_dynamic(test_year, target_variable, pred_months, expected_length)
         if self.process_static:
-            self._process_static()
+            self._process_static(test_year=test_year)
```
tommylees112, line 51: requires `test_year` so that can ignore test years in the calculation
of spatial means and global means of dynamic variables. 

These variables are now calculated from the original data.nc file

src/models/data.py
```diff
@@ -485,7 +500,12 @@ def calculate_static_normalizing_array(
             for norm_var in normalizing_dict_keys:
                 if var == norm_var:
                     mean.append(self.static_normalizing_dict[norm_var]["mean"])
-                    std.append(self.static_normalizing_dict[norm_var]["std"])
+                    # TODO: DO NOT ALLOW STD TO BE ZERO
```
tommylees112, line 503: Is this the correct way to enforce htat std is not 0? Should be be something really small if the variation is low rather than 1.0? .e.g `1e-10` ?

src/models/data.py
```diff
@@ -516,6 +536,7 @@ def calculate_normalizing_array(
         return normalizing_array
 
     def _calculate_aggs(self, x: xr.Dataset) -> np.ndarray:
+        warnings.warn("Deprecated for causing the static data to vary")
```
tommylees112, line 539: I think i want to remove this function right? Stops the static data from being static ..?

src/models/data.py
```diff
@@ -596,8 +617,23 @@ def _calculate_latlons(self, x: xr.Dataset) -> Tuple[np.ndarray, np.ndarray]:
         return latlons, train_latlons
 
     def _calculate_static(self, num_instances: int) -> np.ndarray:
+        # make sure values are DROPPED!
+        assert self.static is not None
+        if self.ignore_vars is not None:
```
tommylees112, line 622: Need to remove the pixel / global means for the ignore_vars from the static data

src/models/data.py
```diff
@@ -611,9 +647,25 @@ def _calculate_static(self, num_instances: int) -> np.ndarray:
                 )
 
                 static_np = (
-                    static_np - self.static_normalizing_array["mean"]
-                ) / self.static_normalizing_array["std"]
+                    (static_np - self.static_normalizing_array["mean"])
+                    / [
+                        s if s != 0 else 1e-10
```
tommylees112, line 652: is this the more correct way to replace the zero std ? I think it should be ONE or the OTHER

src/models/neural_networks/base.py
```diff
@@ -116,6 +118,9 @@ def train(
         # initialize the model
         if self.model is None:
             x_ref, _ = next(iter(train_dataloader))
+            assert (
```
tommylees112, line 121: hmmmm... why am i asserting that there has to exist static data?

src/models/neural_networks/ealstm.py
```diff
@@ -297,8 +301,19 @@ def forward(
             static_x.append(static)
         if self.include_pred_month:
             static_x.append(pred_month)
+
+        # append prev_y to DYNAMIC data
```
tommylees112, line 305: the previous y value is not static so should be appended to the dynamic data instead of the static data

src/models/neural_networks/ealstm.py
```diff
@@ -297,8 +301,19 @@ def forward(
...
         if self.include_prev_y:
-            static_x.append(prev_y)
+            # TODO: Gabi can you check this ?
```
tommylees112, line 307: **IMPORTANT**: is this doing what I think it is re: appending the `prev_y` value to the dynamic data

tests/models/neural_networks/_build_static_data.py
```diff
@@ -0,0 +1,123 @@
+"""
+functions for calculating the static embedding
```
tommylees112, line 2: This is only here for testing purposes while we decide where the best location is.

Do you think it should be a set of functions in ANALYSIS or as methods of the `ealstm` model? The static embedding is unique to the ealstm right so does it make sense there or in a new analysis class?

src/models/data.py
gabrieltseng: If the variance is equal to 0, dividing by a very small number won't change the variance (especially because the values will be equal to 0 when we subtract the mean), so its easier to just divide by 1 and leave it as is.

src/models/data.py
gabrieltseng: Not removed

src/models/data.py
gabrieltseng: I would make the std 1 if it is zero - see the point above

src/models/data.py
gabrieltseng: Not done yet

src/models/data.py
gabrieltseng: Hmm... I see an argument for keeping this. It stops certain dimensions in the static data from being static, but then we can just do `calculate_aggs=False` in the model. On the other hand, I do think instance-wise aggregations could be useful

src/models/neural_networks/base.py
gabrieltseng: good question? have you tried running things without static data? does it break?

src/models/neural_networks/ealstm.py
gabrieltseng: so this appends prev_y to each dynamic datastep.

Do we want to do this? I see why we would want static data to (sometimes) be globally static, but I also think keeping the previous behaviour (at least as an option) is desirable

src/models/data.py
tommylees112: Okay they could just go in the dynamic data instead then?

src/models/data.py
tommylees112: done in the lines below:
```
            include_vars = [
                v
                for v in self.static.data_vars
                if all([ign_v not in v for ign_v in self.ignore_vars])
            ]
            self.static = self.static[include_vars]
```

src/models/neural_networks/base.py
tommylees112: removed

src/models/neural_networks/ealstm.py
tommylees112: why would it be desirable though? Because that data is still being seen by the model?

tommylees112: I do think that 'static' data should remain temporally static. 

The previous behaviour means that 'static' data is only static with regards to that particular target, however, it changes as the target changes, and hance analysing it requires you to analyse it over time.

Resolving this

PR #158, opened 2020-04-12
Batch the predictions, to avoid memory errors

gabrieltseng:
@tommylees112 - I'll try and consolidate the code, but it should work as is


tommylees112:
do you want me to merge it to master or can i just use as is?

